PR #14, opened 2019-11-13
Add deprecation policy

mfocko:
Related to packit-service/ogr#121

I looked into the options suggested by @lachmanfrantisek which were:
- `Deprecated`
  - seems as a good choice, offers decorator that has optional parameters such as version or custom message

- `Python-Deprecated`
  - deprecated `Deprecated`, which is probably kept in PyPI just for backward-compatibility

- `warnings` (built-in module)
  - seems like a lot of copy pasting of the same code, or manual implementation of `@deprecated`

### DIFF
--- a/deprecation/README.md
+++ b/deprecation/README.md
@@ -1,0 +1,5 @@
+# Deprecation policy
+
+* `Deprecated` PyPI package is used for deprecation
+* annotate deprecated functions with `@depracted` decorator, optionally with message
+* functions will be removed after 6 months from the release where they were marked as deprecated
--- a/deprecation/README.md
+++ b/deprecation/README.md
@@ -1,18 +1,49 @@
 # Deprecation policy
 
-* `Deprecated` PyPI package is used for deprecation
-* annotate deprecated functions with `@depracted` decorator, optionally with message
-* functions will be removed after 6 months from the release in which they were marked as deprecated
+- `Deprecated` PyPI package is used for deprecation
+- annotate deprecated functions with `@depracted` decorator, optionally with message
+- functions will be removed after 6 months from the release in which they were marked as deprecated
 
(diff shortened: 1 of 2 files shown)


lachmanfrantisek:
Please, add the description of this PR to the file as well... (It is useful to know the alternatives and why we choose that specific one.)

deprecation/README.md
```diff
@@ -2,4 +2,17 @@
 
 * `Deprecated` PyPI package is used for deprecation
 * annotate deprecated functions with `@depracted` decorator, optionally with message
-* functions will be removed after 6 months from the release where they were marked as deprecated
+* functions will be removed after 6 months from the release in which they were marked as deprecated
```
lachmanfrantisek: The number of months is your suggestion or the info from `Deprecated` module?

deprecation/README.md
```diff
@@ -2,4 +2,17 @@
...
+
+## Choice of library
+
+Looked into the options suggested by @lachmanfrantisek which were:
```
lachmanfrantisek: Feel free to find more..;) (I've done only a quick search.)

deprecation/README.md
```diff
@@ -2,4 +2,17 @@
...
+
+* `Deprecated`
+  * seems as a good choice, offers decorator that has optional parameters such as version or custom message
```
lachmanfrantisek: Some code example?..;)

(You can show here that you spent some time playing with that.)

deprecation/README.md
mfocko: I looked into few projects that have explicit policy...
However thinking about it, it should be more specific, since it could mean deprecating function in 0.X.Y version and then removing it right in the next version after Y.

PR #20, opened 2019-11-21
Specify removal of deprecated functionality

mfocko:
@TomasTomecek like that?
I'm not sure about specifying it in the warning since we could use minor releases, etc.

### DIFF
--- a/deprecation/README.md
+++ b/deprecation/README.md
@@ -2,7 +2,8 @@
 
 - `Deprecated` PyPI package is used for deprecation
 - annotate deprecated functions with `@depracted` decorator, optionally with message
-- functions will be removed after 6 months from the release in which they were marked as deprecated
+- functions will be removed in the 5th release after the release in which they were
+  deprecated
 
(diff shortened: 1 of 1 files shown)


mfocko:
I've added template for deprecation message, I think it can be shorter since the warning looks like:
```
DeprecationWarning: Call to deprecated method {name}. ({custom_message})
```

deprecation/README.md
```diff
@@ -2,7 +2,9 @@
 
 - `Deprecated` PyPI package is used for deprecation
 - annotate deprecated functions with `@depracted` decorator, optionally with message
-- functions will be removed after 6 months from the release in which they were marked as deprecated
+- functions will be removed in the 5th release after the release in which they were
+  deprecated
+  template for warning: `since {version}, will be removed in {version}: {alternative/message}`
```
lachmanfrantisek:
```suggestion
-  template for warning: `since {version}, will be removed in {version}: {alternative/message}`
```

PR #26, opened 2019-12-13
add meeting minutes w/ Dusty

TomasTomecek:


LorbusChris:
@TomasTomecek should we setup a meeting? I'll also be at DevConf.cz again in January :)

LorbusChris:
@TomasTomecek should we setup a meeting? I'll also be at DevConf.cz again in January :)

PR #30, opened 2020-03-03
Create script to convert dist-git to source-git

csomh:
Convert the tip of a branch from a dist-git repo into a commit on the
selected branch in source-git.

### DIFF
--- a/dist-git-to-src-git/README.md
+++ b/dist-git-to-src-git/README.md
@@ -20,7 +20,7 @@ Here are the steps:
 
 3. Now applying the patches (from the `SOURCES` directory) can begin. In order
    to decide which patches to apply, look in the spec-file. Use
-   [rebase-helper] for this.
+   [rebase-helper's `get_applied_patches()`] for this.
 
 ## Intermezzo: what should happen to all the patches?
 
@@ -30,21 +30,20 @@ Here is what we could do with all the patches:
   be deleted, a.k.a they wont show up in the source-git tree.
 
 - Patches which are used in the spec file and cannot be applied - this is
-  where tooling should err out, as something bad happened in dist-git. Or? It
-  might also happen that the patch cannot be applied b/c it was applied
-  already? Will we able to tell this?
+  where tooling should err out, as something bad happened in dist-git.
 
 - Patches which are not used in the spec file will end up being in the
   source-git tree under `centos-packaging/SOURCES/`. Maintainers might want to
   clean this up in the future. Or these might be other kinds of files stored
-  for unknown reasons.
+  for yet unknown reasons.
 
 ### For the future
 
 Are there any expectations for how the history of a source-git repo will
 evolve?
 
-What should happen when sources in dist-git are updated?
+What should happen when sources in dist-git are updated? Currently we will
+recreate the source-git branch.
 
 What should happen when the spec file or the patches in dist-git are updated?
 
@@ -76,3 +75,9 @@ Or breaking it down:
     $ dist2src copy-patches rpms/rpm src/rpm
     $ dist2src apply-patches [--keep-files] src/rpm
     $ dist2src commit src/rpm
+
+`dist2src get-archive` calls [`get_sources.sh`] or the script specified in
+`DIST2SRC_GET_SOURCES`.
+
+[`get_sources.sh`]: https://wiki.centos.org/Sources#get_sources.sh_script
+[rebase-helper's `get_applied_patches()`]: https://github.com/rebase-helper/rebase-helper/blob/e98f4f6b14e2ca2e8cbb8a8fbeb6935e5d0cf289/rebasehelper/specfile.py#L351
--- a/dist-git-to-src-git/README.md
+++ b/dist-git-to-src-git/README.md
@@ -1,0 +1,78 @@
+# Converting dist-git to source-git
+
+This should be similar to [How to source-git?], except that it isn't :)
+
+The main difference is that [How to source-git?] starts with the upstream
+history and applies the patches from dist-git.
+
+In this case though, the starting point is dist-git itself, so a few things
+need to be done differently.
+
+Here are the steps:
+
+*Note:* things bellow were done on git.centos.org/rpms/rpm.
+
+1. Fetch the sources from the lookaside cache.
+
+   This will create some `tar.*` under `SOURCES`.
+
+2. Unpack the `tar` to `src/rpm`.
+
+3. Now applying the patches (from the `SOURCES` directory) can begin. In order
+   to decide which patches to apply, look in the spec-file. Use
+   [rebase-helper] for this.
+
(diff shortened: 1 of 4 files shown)


dist-git-to-src-git/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+    $ cd git.centos.org
+    $ dist2src checkout rpms/rpm c8s
+    $ dist2src get-archive rpms/rpm
+    $ dist2src checkout --orphan src/rpm c8s
+    $ dist2src extract-archive rpms/rpm src/rpm
+    $ dist2src copy-spec rpms/rpm src/rpm
+    $ dist2src copy-patches rpms/rpm src/rpm
+    $ dist2src apply-patches [--keep-files] src/rpm
+    $ dist2src commit src/rpm
```
csomh: @TomasTomecek so obviously this is still WIP, but as a first feedback: will these steps lead to the result we want? Is the process okay the way it's described in this README?

dist-git-to-src-git/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+1. Fetch the sources from the lookaside cache.
+
+   This will create some `tar.*` under `SOURCES`.
+
+2. Unpack the `tar` to `src/rpm`.
+
+3. Now applying the patches (from the `SOURCES` directory) can begin. In order
+   to decide which patches to apply, look in the spec-file. Use
+   [rebase-helper] for this.
```
TomasTomecek: why not just do centpg/fedpkg `prep`?

dist-git-to-src-git/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+Here is what we could do with all the patches:
+
+- Patches which are used in the spec file and are applied to the sources will
+  be deleted, a.k.a they wont show up in the source-git tree.
+
+- Patches which are used in the spec file and cannot be applied - this is
+  where tooling should err out, as something bad happened in dist-git. Or? It
+  might also happen that the patch cannot be applied b/c it was applied
+  already? Will we able to tell this?
```
TomasTomecek: such patches wouldn't be in the spec b/c the build would fail - if a patch cannot be applied, it's an error in the spec file

dist-git-to-src-git/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+- Patches which are not used in the spec file will end up being in the
+  source-git tree under `centos-packaging/SOURCES/`. Maintainers might want to
+  clean this up in the future. Or these might be other kinds of files stored
+  for unknown reasons.
+
+### For the future
+
+Are there any expectations for how the history of a source-git repo will
+evolve?
```
TomasTomecek: so far it won't, but ultimately it should grow like any other upstream project git history

dist-git-to-src-git/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+What should happen when sources in dist-git are updated?
```
TomasTomecek: recreate

dist-git-to-src-git/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+What should happen when the spec file or the patches in dist-git are updated?
```
TomasTomecek: recreate

dist-git-to-src-git/README.md
```diff
@@ -0,0 +1,78 @@
...
+    $ dist2src convert rpms/rpm:c8s src/rpm:c8s
+
+Or breaking it down:
+
+    $ cd git.centos.org
+    $ dist2src checkout rpms/rpm c8s
+    $ dist2src get-archive rpms/rpm
+    $ dist2src checkout --orphan src/rpm c8s
+    $ dist2src extract-archive rpms/rpm src/rpm
+    $ dist2src copy-spec rpms/rpm src/rpm
```
TomasTomecek: will these copy all the packaging files? meaning, spec file + everything in %sources

dist-git-to-src-git/README.md
csomh: b/c I wasn't aware of it :)

dist-git-to-src-git/README.md
csomh: currently only spec. `copy-patches` will copy the patches.

dist-git-to-src-git/dist2src.py
```diff
@@ -0,0 +1,181 @@
...
+        options["B"] = branch
+
+    repo.git.checkout(**options)
+
+
+@cli.command()
+@click.argument("path")
+def get_archive(path):
+    # TODO(csomh): make the path to get_sources.sh configurable
+    get_sources = sh.Command("get_sources.sh")
```
jpopelka: Where is the `get_sources.sh`? I don't see it in this PR.
Getting sources should be as easy as `spectool -g *.spec`.

csomh: https://wiki.centos.org/Sources#get_sources.sh_script

csomh: `spectool` downloads from the URL in the spec file. `get_sources.sh` downloads the archive from the lookaside cache as references in the [.metadata](https://git.centos.org/rpms/rpm/blob/c8s/f/.rpm.metadata) file.

dist-git-to-src-git/README.md
TomasTomecek: I can see that `prep` applies also all the patches after it unpacks the archive

I wonder if we need to apply patches in a dedicated step

csomh: Not necessarily, but the patches that are applied need to be removed from the spec file in order to keep things in a working state.

I'll have a look at what exactly `fedpkg prep` does.

In general I agree that it would be better to use existing tooling than to reinvent the wheel.

dist-git-to-src-git/README.md
csomh: True.

dist-git-to-src-git/README.md
TomasTomecek: fedpkg prep is just a wrapper around `rpmbuild -bp`

TomasTomecek: you can also look at what packit srpm does, same concept, different rpmbuild command

csomh: I've looked into all the above. 

While all the methods above would unpack the source and apply the patches, in the case of dist-git to src-git conversion, those patches also need to be removed from the spec file. 

Hooking this into the approaches mentioned above seems to me to be at least as complicated (if not more) than the current implementation (we would still need to get the list of patches applied and comment them out from the SPEC-file).

So if you agree, I would stick to the current implementation, at least, I don't see any solid benefit to change it now. WDYT?

csomh:
A test will be added in a follow up PR.

PR #34, opened 2020-03-13
[dist-git-to-src-git] README & Makefile details

jpopelka:


dist-git-to-src-git/Makefile
```diff
@@ -2,12 +2,14 @@
 
 PACKAGE ?= rpm
 BRANCH ?= c8s
+DIR ?= git.centos.org
```
csomh: The original idea behind 'make check' was to serve as a very basic test for the script. This is why the dir was called `test` and it is also included in `.gitignore`. 

If we do this, then let's remove `.gitignore`, rename the target to `convert` (as in `make convert`) and update the usage string to avoid future confusion.

jpopelka: Ok, PTAL

Issue #33, opened 2020-03-11
[dist-git-to-src-git] TODO


jpopelka:
Just tried `dist-git-to-src-git` and it worked like a charm.
There are however few things that need to be changed/added for `packit` to be able to use the created source-git repo.

I was expecting to see individual commits for each downstream patch, however, I found only one commit - maybe I'm looking at wrong place.

So before we start applying downstream patches we need to:
```bash
$ echo "upstream_ref: c8s-sg-start" > .packit.yaml
$ git add .packit.yaml; git commit -am ".packit.yaml"
$ git tag c8s-sg-start HEAD
for patch in downstream-patches:
  if is git patch:
    git am patch
  else:
    patch -p1 --verbose --fuzz=0 <patch
    git commit -am "fix"
```

I'll update this description tomorrow, this is how I see it now.

csomh:
dist2src does a dist-git commit to source-git commit mapping currently, because it has no knowledge of the upstream history (only the dist-git history). So tracking future dist-git history, that is: updating the source-git history when there is a new commit in dist-git, would be problematic. What do we do with the individual commits applying the downstream patches, for example? How do we deal with them but still have a linear source-git history, instead of just rewriting all?

The ideal long term solution might be to make dist2src aware of the upstream history (i.e. the commit ref corresponding to the archive found in the lookaside cache), and apply the patches on top of that.

We can discuss more in person.

TomasTomecek:
what's the status here?

jpopelka:
Should be done with https://github.com/packit-service/dist-git-to-source-git/pull/2

Closing

PR #36, opened 2020-03-16
Use `git am` to apply patches

csomh:
This way applying patches in git-format will produce more meaningful
commits. Default to `git apply` only when `git am` fails.

Related to #33.


jpopelka:
tested, works OK, merging

PR #37, opened 2020-03-16
Utilize click.Path and pathlib.Path

jpopelka:


dist-git-to-src-git/dist2src.py
```diff
@@ -103,7 +105,7 @@ def checkout(path, branch, orphan=False):
 
 
 @cli.command()
-@click.argument("gitdir")
+@click.argument("gitdir", type=click.Path(exists=True))
```
lachmanfrantisek: We can test more things:

```suggestion
@click.argument("gitdir", type=click.Path(exists=True, dir_okay=True, file_okay=False, readable=True))
```

```
:param exists: if set to true, the file or directory needs to exist for
                   this value to be valid.  If this is not required and a
                   file does indeed not exist, then all further checks are
                   silently skipped.
    :param file_okay: controls if a file is a possible value.
    :param dir_okay: controls if a directory is a possible value.
    :param writable: if true, a writable check is performed.
    :param readable: if true, a readable check is performed.
    :param resolve_path: if this is true, then the path is fully resolved
                         before the value is passed onwards.  This means
                         that it's absolute and symlinks are resolved.  It
                         will not expand a tilde-prefix, as this is
                         supposed to be done by the shell only.
    :param allow_dash: If this is set to `True`, a single dash to indicate
                       standard streams is permitted.
    :param path_type: optionally a string type that should be used to
                      represent the path.  The default is `None` which
                      means the return value will be either bytes or
                      unicode depending on what makes most sense given the
                      input data Click deals with.
```

jpopelka: `dir_okay=True` and `readable=True` are the default
I can set `file_okay=False` (it's `True` by default) if it makes you happy :-)

lachmanfrantisek: (I like using the default ones as well as a kind of documentation.)

dist-git-to-src-git/README.md
```diff
@@ -68,13 +68,13 @@ Or breaking it down:
 
     $ cd git.centos.org
     $ dist2src checkout rpms/rpm c8s
-    $ dist2src get-archive rpms/rpm
```
csomh: Although `extract-archive` will make sure the archive was downloaded, maybe it would make sense to keep it here, just to be explicit.

dist-git-to-src-git/dist2src.py
jpopelka: Now I don't understand your last note.
Do you expect me to apply your suggestion (`dir_okay=True, file_okay=False, readable=True`) because it's better to be explicit (kind of documentation)?
Or do you just want me to add some kind of documentation? What kind? :-)

lachmanfrantisek: That was only my personal preference for using these values explicitly as a form of documentation.

My prefered way would be `dir_okay=True, file_okay=False, readable=True`, but `file_okay=False` is fine.

jpopelka:
All fixed, PTAL @lachmanfrantisek @csomh

PR #38, opened 2020-03-19
WIP: Support https://github.com/facebookincubator/rpm-backports

jpopelka:


jpopelka:
Uh, I took it from the wrong end. We won't be creating source-git repos from the FB repository, but from centos dist-git and only apply FB patches in a fork.

PR #41, opened 2020-04-06
external testing research

TomasTomecek:


external-testing/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+| A feature | VCRpy | requre |
+|-----------|-------|--------|
+| Record & replay for python requests | ✓ | ✓ |
+| Match on call stack | ✘ | ✓ |
+| Match on URL & request type | ✓ | ✘ |
+| Record & replay for arbitrary functions | ✘ | ✓ |
+| Filtering sensitive data out of recorded responses | ✓ [1] | ✓ [2] |
+| Easy to use | ✓ [3] | ✘ [4] |
+| We control the project | ✘ | ✓ |
```
lachmanfrantisek, line 44: Can you please try the use-case with multiple calls to the same API? E.g. test for creating new issue:
- list all issues (1.)
- create issue
- list all issues (2.)

Same call but with a different response.

external-testing/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+* [1] https://vcrpy.readthedocs.io/en/latest/advanced.html#filter-sensitive-data-from-the-request
+* [2] https://requre.readthedocs.io/en/latest/usages/postprocessing.html
+* [3] [You can use a decorator or a context manager](https://vcrpy.readthedocs.io/en/latest/usage.html)
+* [4] [Tenths of lines of code](https://github.com/packit-service/packit/blob/master/tests_recording/__init__.py#L16) [which are very hard to understand](https://github.com/packit-service/packit/blob/master/tests_recording/testbase.py#L22)
+
+### TL;DR
+
+These are my personal (@TomasTomecek) comments and opinions:
```
lachmanfrantisek, line 53: This should be an objective review. You preconceive the result...;)

external-testing/README.md
```diff
@@ -0,0 +1,60 @@
...
+mean? Well, talking to GitHub API, COPR API, running `fedpkg`, cloning
+repositories...
+
+
+## Use cases
+
+* Interact with an external API (such as GitHub, COPR, Pagure, FAS, k8s)
+  * Authenticated vs. unauthenticated calls.
+* Clone a git repository.
+* Construct a git repository
```
lachmanfrantisek, line 13: What about push -- we can use it for creating PRs that are part of multiple jobs.

external-testing/README.md
lachmanfrantisek: Another use-case:
- Different versions of library we use (e.g. `rebase-helper`, `pygithub`, `gitlab-python`, `copr`, ...) have a different set of calls. Can vrcpy handle this?

external-testing/README.md
```diff
@@ -0,0 +1,60 @@
...
+* VCRpy
+
+
+## VCRpy vs. requre
+
+The section you are craving to read. It's opinionated and not objective.
+
+| A feature | VCRpy | requre |
+|-----------|-------|--------|
+| Record & replay for python requests | ✓ | ✓ |
```
lachmanfrantisek, line 38: Compatibility with `vcr` cassettes? Using some kind of standard can be beneficial. (If there is be some tool e.g. for postprocessing/visualisation/... it will probably be for some standard format.)

[Betamax](https://github.com/betamaxpy/betamax) should be compatible. Not sure about `vcrpy`.

external-testing/README.md
TomasTomecek: ad second comment, that should be handled here: https://github.com/packit-service/packit/pull/785

ad first: I'll create a test in ogr, let me know if you have ideas what to test

external-testing/README.md
TomasTomecek: but I'm really objective in the table, just the tl;dr section is subjective, will update

external-testing/README.md
TomasTomecek: I'd say push is covered by: run external command && R/W actions on a remote service

external-testing/README.md
TomasTomecek: what?

lachmanfrantisek:
```python
if what == "supprise":
     return "Just read it in the betamax readme. If there are multiple recording tools that support the same format, we can switch between them more easily."
elif what == "I don't understand.":
    return "The response format can be shared between multiple implementations. E.g. with original ruby `vcr`."
else:
    return "What do you mean?"
```

lachmanfrantisek:
Other related projets:
- https://github.com/agriffis/vcrpy-unittest (And the related [issue](https://github.com/kevin1024/vcrpy/issues/512) to incorporate it directly to the vcrpy  that btw mentions the end of maintenance of the original author.)
- https://github.com/ktosiek/pytest-vcr
- https://github.com/kiwicom/pytest-recording

lachmanfrantisek:
Is vcrpy somehow extendable to implement custom requests? (I am thinking about possibility of making requre just an extension of the vcrpy.)

TomasTomecek:
What's a custom request?

external-testing/README.md
lachmanfrantisek: https://github.com/orgs/betamaxpy/people if we want some redhater maintaining the project...;)

lachmanfrantisek:
> What's a custom request?

Like the external command, our function, ... (Just to implement only missing pieces of vcrpy.)

external-testing/README.md
TomasTomecek: they are simple yaml files, even if they deviated from the standard, we should be able to transform them - I'd say it should be easily feasible to e.g. transform requre response files to vcr

external-testing/README.md
lachmanfrantisek: Maybe try the `create_issue` as I wrote -- I am interested in a situation where you have exactly the same call with two different responses.

In other words, how the vcrpy differentiates the two requests? (In you example, they differ a bit.)

TomasTomecek: I did with the whoami call - vcrpy uses matching on the request, so it's not just url, it's also headers, params and data.

external-testing/README.md
```diff
@@ -0,0 +1,65 @@
...
+| Match on call stack | ✘ | ✓ |
+| Match on URL & request type | ✓ | ✘ |
+| Calling the same URL twice works as expected | ✓ | ✘ |
+| Record & replay for arbitrary functions | ✘ | ✓ |
+| Filtering sensitive data out of recorded responses | ✓ [1] | ✓ [2] |
+| Easy to use | ✓ [3] | ✘ [4] |
+| Actively maintained | ✘ | ✓ |
```
csomh, line 45: What does this mean? I see that VCRpy is [looking for maintainers](https://github.com/kevin1024/vcrpy/issues/469), but does that meant that is not actively maintained?

external-testing/README.md
lachmanfrantisek: Yes, but what if we want exactly same request, like this test:
1. get all issues (returns 42 issue objects)
2. create a new issue
3. get all issues  (returns 43 issue objects) -- same URL, same headers, same params and same data...

external-testing/README.md
lachmanfrantisek: I would not say "unmaintained", but we don't know. Here is a response in the issue I linked above:

> Hi, I just wanted to update active PRs and issues to let you know I will no longer be an active maintainer on this project.
>
> I recently changed jobs where I no longer use VCRpy or Python at all so I don't have the bandwidth anymore. I won't be monitoring issues or PRs either.
>
> I will have to delegate to other maintainers or ask for some contributors to step up.
> 
> I know the Azure CLI uses VCRpy so it would be nice if they could spare some sprint time to maintain this project as it was invaluable in our CI when I was using it.
>
> Kind regards,
> Josh

TomasTomecek: `Latest commit accffa8 on 20 Dec 2019` and the author clearly states he's no longer using it nor interested in working on it - that's a definition of a no longer actively maintained project for me

maybe betamax would be a better pick

TomasTomecek:
> Would it make sense to have a relatively detailed comparison of what work would be required to switch to VCRpy, or improve requre as is, or improve requre by adopting VCR, or (something else)? I think something like this would help us to reach a decision which everyone can commit to, and would help us have a clear roadmap for the future.

I personally don't think we need such analysis - we just need to pick a solution, as a team, and stick to it. To me it doesn't matter if it takes 3 or 21 story points to achieve it, instead we need a solution which works (ideally) for everyone and we're happy with it as a team.

csomh:
> I personally don't think we need such analysis - we just need to pick a solution, as a team, and stick to it. To me it doesn't matter if it takes 3 or 21 story points to achieve it, instead we need a solution which works (ideally) for everyone and we're happy with it as a team.

Given that the team has limited bandwidth I would strongly oppose this: I think it's very important to understand the effort required to follow one or the other path, and have a good understanding what does it take to reach a good working state.

This work (improving requre, switching to VCRpy, whatever) will compete with other work we do. If it takes 21 story points, but we'll never have the bandwidth to spend those 21 story points because of other priorities, we might end up to be where we are today. Compared to this, a small 3 story-point improvement which has a higher chance to be completed is much better.

external-testing/README.md
csomh: Okay, I missed this.

Isn't betmax requests-only?

TomasTomecek:
> > I personally don't think we need such analysis - we just need to pick a solution, as a team, and stick to it. To me it doesn't matter if it takes 3 or 21 story points to achieve it, instead we need a solution which works (ideally) for everyone and we're happy with it as a team.
> 
> Given that the team has limited bandwidth I would strongly oppose this: I think it's very important to understand the effort required to follow one or the other path, and have a good understanding what does it take to reach a good working state.
> 
> This work (improving requre, switching to VCRpy, whatever) will compete with other work we do. If it takes 21 story points, but we'll never have the bandwidth to spend those 21 story points because of other priorities, we might end up to be where we are today. Compared to this, a small 3 story-point improvement which has a higher chance to be completed is much better.

Let's talk tomorrow on the standup - doesn't make sense to argue here :)

external-testing/README.md
lachmanfrantisek: Do we use (internally) something different that requests? (It's the main difference between vcrpy and betamax.)

TomasTomecek: we are using requests only afaik (copr client is using requests as well); status command is being done in async way but not with asyncio to my knowledge

external-testing/README.md
TomasTomecek: I can try.

external-testing/README.md
```diff
@@ -0,0 +1,65 @@
...
+* Betamax
+
+
+## VCRpy vs. requre
+
+| A feature | VCRpy | requre |
+|-----------|-------|--------|
+| Record & replay for python requests | ✓ | ✓ |
+| Match on call stack | ✘ | ✓ |
+| Match on URL & request type | ✓ | ✘ |
```
jscotka, line 40: it is possible also in requre (but it is harder to use, but could be improved to store url as well), because requre is more generic than just to `request` module. 
In fact `requre` allows you to store any parameter of function as key (url) as well, but we have to ensure to remove sensitive data like login and password from url

external-testing/README.md
```diff
@@ -0,0 +1,65 @@
...
+| Calling the same URL twice works as expected | ✓ | ✘ |
```
jscotka, line 41: it is also possible. it was the first implementation what were in `OGR` it used `URL` and `data` keys for storing responses.
It is also possible now, but it is not the default mode. it were changed to use list by default, instead of just dict, what matched results.

external-testing/README.md
```diff
@@ -0,0 +1,65 @@
...
+| Record & replay for arbitrary functions | ✘ | ✓ |
+| Filtering sensitive data out of recorded responses | ✓ [1] | ✓ [2] |
+| Easy to use | ✓ [3] | ✘ [4] |
```
jscotka, line 44: depends :-) jup, vrcpy is easier for easier testcases, when you can decorate code directly for requrest module and match specific URL to record, but not in case when you would like to work transparently to python execution.

external-testing/README.md
```diff
@@ -0,0 +1,65 @@
...
+| Actively maintained | ✘ | ✓ |
+
+* [1] https://vcrpy.readthedocs.io/en/latest/advanced.html#filter-sensitive-data-from-the-request
+* [2] https://requre.readthedocs.io/en/latest/usages/postprocessing.html
+* [3] [You can use a decorator or a context manager](https://vcrpy.readthedocs.io/en/latest/usage.html)
+* [4] [Tenths of lines of code](https://github.com/packit-service/packit/blob/master/tests_recording/__init__.py#L16) [which are very hard to understand](https://github.com/packit-service/packit/blob/master/tests_recording/testbase.py#L22)
```
jscotka, line 50: these tenth of lines are caused to be able to work more generic, in case you would like to record just requests module and nothing else:
 ```
upgraded_import_system = (upgrade_import_system().decorate(where="^requests$", what="Session.send", decorator=RequestResponseHandling.decorator(item_list=[])))
```
exactly stores all requests to `request` module.

File: external-testing/README.md [hunk 10]
```diff
@@ -0,0 +1,65 @@
...
+These are my personal (@TomasTomecek) comments and opinions:
+* VCRpy's API and documentation is focused on the record & replay use case
+  which makes it very pleasant to work with.
+* requre's documentation and API is centered around implementation details
+  which makes it very configurable, but unfortunately extremely hard to work
+  with - even to the point that you need to understand internals.
+* VCRpy in action:
+  * [Basic usage](https://github.com/packit-service/packit/pull/785)
+  * [Calling the same URL twice](https://github.com/packit-service/ogr/pull/373)
+* The fact that requre is matching replay via call stack makes it less flexible
```
jscotka, line 63: stack info are there important to be able to differenciate to store and replay various functions in generic way. in case focusing just to request module, then it could be simplier, and also there is there is SimpleKeyStorage, what stores just function and his module, and supress all other stack info to be important as an minimal match. But the trouble is, that in that case you have to solve very well which keys to use then (arguments of functions) or wrote own wrappers for specific functuions what will solve internal structure of data. (eg similar to recording function from `git` module what will store Fetch/PushInfo) objects to files

external-testing/README.md
```diff
@@ -0,0 +1,65 @@
...
+  and will require data regeneration more often.
+* In my opinion: VCRpy masters its job very well.
```
jscotka, line 65: right, as it is focused just to specific python module, it could be more precise and easier to use for this specific case.

external-testing/README.md
```diff
@@ -0,0 +1,65 @@
...
+* [4] [Tenths of lines of code](https://github.com/packit-service/packit/blob/master/tests_recording/__init__.py#L16) [which are very hard to understand](https://github.com/packit-service/packit/blob/master/tests_recording/testbase.py#L22)
+
+### TL;DR
+
+These are my personal (@TomasTomecek) comments and opinions:
+* VCRpy's API and documentation is focused on the record & replay use case
+  which makes it very pleasant to work with.
+* requre's documentation and API is centered around implementation details
+  which makes it very configurable, but unfortunately extremely hard to work
+  with - even to the point that you need to understand internals.
```
jscotka, line 59: right, documentation should be improved for consumers. question is, if in case requre try to be more generic, if it is able to document it without internals in so easy way.
but maybe examples like in presentation we've done, could be incorporated to requre doc as well. there are described easier way how to use it that in oficial doc

external-testing/README.md
TomasTomecek: I was able to start using vcrpy in minutes. With requre, it took me tenths of minutes and in the end I had to ask you for help. I'm sorry, no, requre is not easy to use.

external-testing/README.md
TomasTomecek: sacrificing user experience over configurability

external-testing/README.md
lachmanfrantisek: Yes, @jscotka is right. It was common in OGR to save a list of responses (exactly the situation I was asking Tomáš) so we made it the default.

TomasTomecek:
@jscotka thank you for the comments - I understand that everything can be changed. This doc focuses on current state which will lay down a base for the next steps we want to take on testing with external services. It wasn't meant to criticize anything and anyone: I have tried to objectively (and then subjectively) compare existing solutions.

Let's discuss tomorrow on standup.

jscotka:
> @jscotka thank you for the comments - I understand that everything can be changed. This doc focuses on current state which will lay down a base for the next steps we want to take on testing with external services. It wasn't meant to criticize anything and anyone: I have tried to objectively (and then subjectively) compare existing solutions.
> 
> Let's discuss tomorrow on standup.

No no, I take it as very useful criticize, not personally :-)
my answers should be also just as an explanation for your points, what are easier for me as an author of the code, where it is more subjective or not easy to undestand how to use it, and not well documented in requre :-)

external-testing/README.md
lachmanfrantisek: This is a solvable problem. Things that will don't take much time and will help a lot:
- renaming of the functions and  arguments
- fixtures for common replacements (e.g. requests)
- add some basic examples to docs

lachmanfrantisek:
I just came across [this vcrpy issue](https://github.com/kevin1024/vcrpy/issues/227#issuecomment-611038185) and it looks like we should try ogr's implementation for GitHub to be sure it works with https.

TomasTomecek:
> I just came across [this vcrpy issue](https://github.com/kevin1024/vcrpy/issues/227#issuecomment-611038185) and it looks like we should try ogr's implementation for GitHub to be sure it works with https.

good point, just tested it and works as expected - the payload is stored as `!!binary`

external-testing/README.md
```diff
@@ -0,0 +1,138 @@
...
+
+#### b) VCRpy + flexmock + pytest fixtures
+
+This implies ditching requre completely. We'd need to:
+* rewrite existing tests to use VCR instead of requre (easy)
+* regenerate or transform data (should be easy)
+* use flexmock (or something else) for git, fedpkg...
+* most importantly: prepare the structure well so it's not messy
+
+This should be achievable in 2 sprints.
```
lachmanfrantisek, line 124: I would consider this as a very optimistic if we want to make it well. (Weeks of polishing is applicable to any solution.)

external-testing/README.md
```diff
@@ -0,0 +1,138 @@
...
+
+Technical arguments end here and personal preference would follow.
+
+
+### Not a solution 3: VCRpy + requre
+
+They don't play well together. This solution doesn't make any sense.
```
lachmanfrantisek, line 131: Have you tried also avoiding requre touching the `requests`? I am not sure how deeply you were trying that...

external-testing/README.md
```diff
@@ -0,0 +1,138 @@
...
+a team discussion to pick one.
+
+
+### Solution 1: Keep using requre and improve it
+
+We can keep using requre as we are now. But given that VCRpy has
+* very neat user experience
+* and matches requests based on their data and not the call stack,
+
+it is very compelling to have VCRpy's experience. This means that if we stick
```
lachmanfrantisek, line 94: That would be a good challenge!

I believe something like requre should exist for Python and would be sad if it dies. 
(I am not saying that should be the main benefit of that solution...;-)

PR #43, opened 2020-05-06
Dashboard Stack

IceWreck:
Paging @csomh 

Link for easier reading. https://github.com/IceWreck/research/blob/master/dashboard/README.md

### DIFF
--- a/dashboard/README.md
+++ b/dashboard/README.md
@@ -99,3 +99,43 @@ podman run -d -p 8300:3000 -u="root"  --name=grafana -v /home/icewreck/Developme
 * Tests ?
 * Adapting it for our use case will be a literal hackfest with tons of plugins thrown in to make it work.
 
+
+
+## Dashboard - List of Features
+
+### Home Page
+
+* Number of successful and failed builds. (bar chart with dual bars for successful/failed) (toggle for weekly/monthly/yearly)
+* Number of successful out of total builds of all time. (donut chart)
+* Total number of projects (100% filled donut chart)
+* Total number of installations (100% filled donut chart)
+* Manually triggered vs automatic builds. (line chart) (toggle for weekly/monthly/yearly)
+* Top 5 most active projects. (Data List) (toggle for weekly/monthly/all-time)
+* Testing farm usage chart
+
+### Build Info Page
+
+* 
+
+### Projects Page
+
+* Total number of projects (100% filled donut chart)
+* Small gallery style cards for every project. (filters to sort by most builds, user namespace, recent activity)
+
+#### Specific Project View
+
+* Project ID
+* Total Number of PRs that used packit-service
+* Total number of builds
+* List of all builds related to that projects, from newest to oldest. (data list)
+* Each build will open the detailed build view
+
+### Status
+
+* Packit Service status and downtime reports.
+* Total number of calls to the GitHub API per hour. (sparkline chart)
+* Number of fedmsgs per hour. (sparkline chart)
+
+### FAQ
+
+### About
--- a/dashboard/README.md
+++ b/dashboard/README.md
@@ -111,11 +111,34 @@ podman run -d -p 8300:3000 -u="root"  --name=grafana -v /home/icewreck/Developme
 * Total number of installations (100% filled donut chart)
 * Manually triggered vs automatic builds. (line chart) (toggle for weekly/monthly/yearly)
 * Top 5 most active projects. (Data List) (toggle for weekly/monthly/all-time)
+* Builds per chroot (donut chart)
 * Testing farm usage chart
 
 ### Build Info Page
 
-* 
+* Package/Project Name
+* SRPM Name
+* Repository URL
+* Link to trigger (pull request or release)
+* Copr Build ID
+* PR ID (where applicable)
+* Project ID
+* Build Submitted, Started, Finished Time
+* List of chroots
+* Status per chroot
+* Instructions to quick enable and install that build
+* List + download link of built RPMS, their size, etc
+* Link to the testing farm results page.
+* Link to the logs page.
+
+#### Build Logs Page
+
+* Once a build has finished, copr creates a web directory listing with downloadable build related files like some log files, the built rpms, etc. 
+* The build logs are in text files like build.log.gz and builder-live.log.gz 
+* These files will be rendered on the logs page inside some sort of monospace text widget.
+* SRPM logs, build logs and the live builder log (tabs to switch between them).
+* Dropdown to switch between the log pages of chroots/targets of the same build id.
+
 
 ### Projects Page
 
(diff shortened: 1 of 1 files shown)


dashboard/README.md
```diff
@@ -1,2 +1,141 @@
...
+* SCSS is a superset of CSS with variable support and better query selection. It is compiled to _readable_ CSS.
+
+### UI/UX
+
+Patternfly has two official variants.
+
+* Patternfly HTML: Its a CSS Framework. Zero inbuilt support for JavaScript. User has to take care of interactive components + animations themselves.
+* Patternfly React: Complete UI toolkit. Components are interactive by default. Inbuilt support for charts and datatables, which follow a similar design pattern.
+
+Note: They also have a slack for community support.
```
TomasTomecek, line 18: ...and the core devs are working for Red Hat :)

dashboard/README.md
```diff
@@ -1,2 +1,141 @@
...
+
+## UI
+
+### Utils
+
+* SCSS is a superset of CSS with variable support and better query selection. It is compiled to _readable_ CSS.
+
+### UI/UX
+
+Patternfly has two official variants.
```
TomasTomecek, line 13: [cockpit project is using patternfly](https://github.com/cockpit-project/cockpit/tree/master/pkg) if you wanna get inspired

dashboard/README.md
```diff
@@ -1,2 +1,141 @@
...
+Since the dashboard will fetch all data from the API, we can make it entirely client side. Using Fetch/Axios, the browser will fetch the json directly from the API, no middleman involved. Controlling the routes in the URL is possible beacuse react-router uses the HTML5 History API. After compiling the JSX to JS, we will have static HTML/CSS/JS which can be served from apache, nginx etc or even GitHub pages.
+
+#### Pros
+* Serving static files is faster than wsgi + python. 
+* One language for everything.
+* Single page web app.
+* Using React + Patternfly's extra React only features like charts and datatables is leaner than importing Patternfly HTML + jQuery + some charts library + datatables library + other plugins.
+
+
+#### Cons
```
TomasTomecek, line 37: do we need to tailor the API jsons for react?

dashboard/README.md
```diff
@@ -1,2 +1,141 @@
...
+* Number of successful out of total builds of all time. (donut chart)
+* Total number of projects (100% filled donut chart)
+* Total number of installations (100% filled donut chart)
+* Manually triggered vs automatic builds. (line chart) (toggle for weekly/monthly/yearly)
+* Top 5 most active projects. (Data List) (toggle for weekly/monthly/all-time)
+* Testing farm usage chart
+
+### Build Info Page
+
+*
```
TomasTomecek, line 118: package name, project, SRPM name, a list of built RPMs, size, link to trigger (PR, release)

dashboard/README.md
IceWreck: Yes, that is where I first saw patternfly.

dashboard/README.md
IceWreck: No, we'll just fetch the json and then parse it as a regular javascipt object in the dashboard

dashboard/README.md
IceWreck: Yes, sorry. This PR was meant to be work in progress. I'll fill the rest of the pages later today

TomasTomecek: no worries, I was really just suggesting

csomh:
@IceWreck please also take a look at #46. We have a possibility to use an internal Prometheus and Grafana instance, so we wouldn't need to maintain them ourselves. 

I see Prometheus as working in combination with the Dashboard.

It would help us working around both the fact that visualising JSON with Grafana is not nice and that Prometheus doesn't play nicely with high-cardinality data. The dashboard would be there to give us a nice view into what Packit Service "knows", while we could use the Prometheus stack for what it is intended: monitoring.

We would count on you to work on the Dashboard and we could take care of Prometheus.

What do you think? Ping me tomorrow please and let's have a chat about this :slightly_smiling_face:

dashboard/README.md
```diff
@@ -1,2 +1,173 @@
...
+* SRPM Name
+* Repository URL
+* Link to trigger (pull request or release)
+* Copr Build ID
+* PR ID (where applicable)
+* Project ID
+* Build Submitted, Started, Finished Time
+* List of chroots
+* Status per chroot
+* Instructions to quick enable and install that build
```
IceWreck, line 129: Lets say the direct install instructions for a package are:
```
$ dnf copr enable icewreck/some-package-10
$ dnf install icewreck/some-package-10
```

Here the _'-10'_ is the GitHub PR ID according to the way packit names copr packages.

But if we do _/packit build_ manually on PR#10 then a completely new build, with a new build id is created. However the install instructions remain the same. So what should we do for the older builds of the same package? Remove direct install instructions for them?

csomh: With these install instructions it's always the package from the last build is going to be installed. So it's fine for the instructions to stay the same.

I *think* packages from previous builds are still available, but one will need to specify the full version when installing. See for example [this PR](https://copr.fedorainfracloud.org/coprs/packit/beaker-project-beaker-22/builds/) which had multiple builds.

csomh:
This might open a somewhat different direction to the current plan presented here, but I'll write it down nevertheless, and let me know @IceWreck what do you think :)

On Friday I did a few thought experiments about how I would use the Dashboard to help understanding and debugging the errors we have in the service.

One of the things that usually happens is that one of our users reports that something went wrong on on of their PRs. Given the project name and the PR ID, I'll start tracking down the issue.

If I'm lucky, the thing that went wrong ended up as an event in Sentry, where I'll have some context provided. But if this is not the case, I'll have to go to OpenShift, and hope that the pods were not redeployed since the issue happened and I'm still able to filter out some of the relevant logs to understand what Packit did for the given PR.

Or I could just craft a smart SQL query to try to extract some information about that PR from the database.

But none of the solutions above is very smart or convenient.

I would rather go to a page, presenting the list of jobs Packit did, in reverse chronological order. Here by job I mean the thing we define in under the [jobs](https://packit.dev/docs/configuration/#jobs) section in `packit.yaml`. So a PR could have multiple jobs in this list.

The list would identify jobs by project and PR id, would present the trigger (ex: `pull_request`) and the type of the job (ex: `copr_build`).

I could also see how the tasks within the jobs are doing: for example, how many chroots are building, finished building successfully or failed.

Selecting an entry would take me to a job page, with more details about the tasks (builds, tests) executed within that job. A could get the link for the log page from there, for example.

The list of jobs could be filtered using a simple search box. I could filter for jobs which run yesterday or last week. I could filter for projects, job types, triggers and so on.

The above has no charts, that's true, but a list like this would be immediately useful in helping us to understand what's going on in Packit, and as the data would be already gathered, we could easily chart this information as a follow up step.

dashboard/README.md
```diff
@@ -1,2 +1,173 @@
...
+* Help menu, login button, settings icon, documentation buttons cannot be removed.
+* Large, publically visible buttons point to grafana documentation, which will be misleading in packit-service's dashboard.
+* We cannot control the URL scheme.
+* I haven't found a way to do something like recursive json.  Eg: (GET and display all builds info and if a user a wants it, find build_id from the all builds json and then do a GET request to fetch another json file for detailed info regarding that build_id)
+* Tests ?
+* Adapting it for our use case will be a literal hackfest with tons of plugins thrown in to make it work.
+
+
+
+## Dashboard - List of Features
```
csomh, line 104: @IceWreck please update this section with what we discussed yesterday, so we can merge this :)

IceWreck: Done

PR #44, opened 2020-05-14
Gitlab Integration

shreyaspapi:


csomh:
How does an Integration become available in a GitLab instance?

shreyaspapi:
> How does an Integration become available in a GitLab instance?

The integration / project service can be made available by adding [app/models/project_services](https://gitlab.com/gitlab-org/gitlab/tree/master/app/models/project_services) - [Docs](https://docs.gitlab.com/ee/user/project/integrations/overview.html#contributing-to-integrations) with adding a new service [here](https://gitlab.com/gitlab-org/gitlab/tree/master/app/models/project_services).

csomh:
> The integration / project service can be made available by adding [app/models/project_services](https://gitlab.com/gitlab-org/gitlab/tree/master/app/models/project_services) - [Docs](https://docs.gitlab.com/ee/user/project/integrations/overview.html#contributing-to-integrations)

So than the Packit integration would become part of the official GitLab distribution? That's a nice way to reach out to a wide audience. Though I think we would still like to support webhook integration.

shreyaspapi:
> So than the Packit integration would become part of the official GitLab distribution? That's a nice way to reach out to a wide audience. Though I think we would still like to support webhook integration.

Hey @csomh yes we will be still supporting the webhook integration as all the events made by webhook or project integration will be the same. I have updated the readme to make it more clear. 

Easy access URL -https://github.com/shreyaspapi/research/blob/master/gitlab_integration/README.md

gitlab_integration/README.md
```diff
@@ -1,16 +1,34 @@
...
 
+## Project integrations
+
+It allow you to integrate GitLab with other applications. They are a bit like plugins in that they allow a lot of freedom in adding functionality to GitLab.
+
+Just like Github Apps but with some limited functionality. This will be a very easy way for any projects to use the Packit service (eg in https://gitlab.gnome.org)
+
+So just by clicking one button, packit service will be active for the project - 
+
+![packit service example](./gitlab-packit.png)
```
lachmanfrantisek, line 11: This is a customizable settings page, isn't it? (So a gui version of our config file.)

Can we react on a change? (So we can e.g. create a PR with the config change or add a badge,...)

gitlab_integration/README.md
```diff
@@ -1,16 +1,34 @@
...
+- In the later stages we can setup a settings page for the packit-service in gitlab which will act as a config for the service instead of tweeking packit config file in the repository.
 
 ### Cons
 
-- Users still have to define auth for hooks (Packit).
-- Users will still have to give access to a packit service user for getting the build updates.
+- Users will still have to give access to a packit service "user" for getting the build updates.
+ - This introduces a new programming language "Ruby" that we will need to work on for maintaining the service in gitlab.
+
+For adding the service to gitlab we will have to contribute to the https://gitlab.com/gitlab-org/gitlab which has to be approved by the admins (Currently looking more into it).
```
lachmanfrantisek, line 31: I would make a specific section for the implementation. (So we can clearly say if it is worth spending the time on it.)

gitlab_integration/README.md
```diff
@@ -1,16 +1,34 @@
...
 ### Cons
 
-- Users still have to define auth for hooks (Packit).
-- Users will still have to give access to a packit service user for getting the build updates.
+- Users will still have to give access to a packit service "user" for getting the build updates.
+ - This introduces a new programming language "Ruby" that we will need to work on for maintaining the service in gitlab.
+
+For adding the service to gitlab we will have to contribute to the https://gitlab.com/gitlab-org/gitlab which has to be approved by the admins (Currently looking more into it). 
+
+All the events created by integration / webhooks are the same so we can even take this up later.
```
lachmanfrantisek, line 33: Relevant point!

In both version, how can the user specify events that will be sent to us?

lachmanfrantisek: I heard that there is a way to setup webhooks globally (automatically) for all projects of the instance. (Can you please try to confirm that? It can help us a lot in case of Fedora-friendly instances like gitlab.gnome.org/.)

But that's only a shortcut and not doable for instances like gitlab.com.

gitlab_integration/README.md
shreyaspapi: okay!

gitlab_integration/README.md
shreyaspapi: >This is a customizable settings page, isn't it? (So a gui version of our config file.)

Yes

>Can we react on a change? (So we can e.g. create a PR with the config change or add a badge,...)

What do you mean by creating a PR with the config change?

gitlab_integration/README.md
shreyaspapi: >In both versions, how can the user specify events that will be sent to us?

As you can see https://github.com/shreyaspapi/research/raw/master/gitlab_integration/gitlab-packit.png Users have trigger options for Push, Issue, Confidential Issue, Merge Request. Only if the option is ticked the events will be sent to us.

shreyaspapi: > I heard that there is a way to setup webhooks globally (automatically) for all projects of the instance. (Can you please try to confirm that? It can help us a lot in case of Fedora-friendly instances like gitlab.gnome.org/.)
> 
> But that's only a shortcut and not doable for instances like gitlab.com.

Even if we set up webhook globally for gnome will they be comfortable for sending all the events for all the projects they have? 

If we do this I am not sure if we can set up trigger options separately for each project.

gitlab_integration/README.md
lachmanfrantisek: That was just an example reaction from our side when the user saves the settings.
- Get the values from the settings page and update the config-file with those values and create a merge-request with the changes.
- It can be misleading to users when I am thinking about that. (They can be in a weird situation when they don't merge the merge-request.)

We can maybe find a better use-case.

gitlab_integration/README.md
lachmanfrantisek: > As you can see https://github.com/shreyaspapi/research/raw/master/gitlab_integration/gitlab-packit.png Users have trigger options for Push, Issue, Confidential Issue, Merge Request. Only if the option is ticked the events will be sent to us.

I am now confused about what values are in the settings-page by default and what can be set by us. So the example in the image shows the default values and we can add custom ones?

lachmanfrantisek: > Even if we set up webhook globally for gnome will they be comfortable for sending all the events for all the projects they have?

Yes, that's a valid concern. We can have some opt-in value in the settings, but in that case, we still need to process all those events.

shreyaspapi: > I am now confused about what values are in the settings-page by default and what can be set by us. So the example in the image shows the default values and we can add custom ones?

So just to make everything clear, There are no default values when adding project services! We can add or remove any type of parameters we want (this is where we can have all the config settings we thought of).

We just have to find alternative to this -

Even after installing to project, the service has no permissions to the repository. So for sending out the commit-status (Build information), a user will still have to give permissions to **Packit Gitlab User**

gitlab_integration/README.md
shreyaspapi: So how about getting the information in the events itself and ditching the config file completely. 

There should be a way when user updates the settings page we can send us an event with updated settings, But for this, we will have to work more on the GitLab side.

gitlab_integration/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+- Users dont have to setup webhook for each project.
+- In the later stages we can setup a settings page for the packit-service in gitlab which will act as a config for the service instead of tweeking packit config file in the repository.
+
+### Cons
+
+- Users will still have to give access to a packit service "user" for getting the build updates.
+ - This introduces a new programming language "Ruby" that we will need to work on for maintaining the service in gitlab.
+
+For adding the service to gitlab we will have to contribute to the https://gitlab.com/gitlab-org/gitlab which has to be approved by the admins (Currently looking more into it).
```
TomasTomecek, line 31: I'm really confused how this works: so we need to contribute the code of the integration service inside official gitlab project? where and how does the service run then?

shreyaspapi: >so we need to contribute the code of the integration service inside official gitlab project?

Yes, we need to contribute code to official GitLab project. 

> where and how does the service run then?

As mentioned at the start it is just like a plugin to a project like [webhook](https://docs.gitlab.com/ee/user/project/integrations/webhooks.html), which will send GitLab events to us ([Packit-service](https://github.com/packit-service/packit-service)).

gitlab_integration/README.md
```diff
@@ -0,0 +1,76 @@
+# Gitlab integration research
+
+## Gitlab Test Instance
+
+[http://52.183.132.26:3000/](http://52.183.132.26:3000/)
+
+### Crediantials
+
+- Username - testpackit
+- Password - Pass@123
```
TomasTomecek, line 10: even though it's a test instance which I assume you can put down any time, you're putting your host in a terrible risk

lachmanfrantisek: Yes, we can share the tokens in a private way.

shreyaspapi: I had put many restrictions before sharing it here. But yeah makes sense.

lachmanfrantisek:
@shreyaspapi Can you please rebase so we can merge it? It would be nice to have it included in the structure.

Thanks!

lachmanfrantisek:
@mfocko Thanks, fell out of my radar...

PR #3, opened 2019-03-18
WIP: Automation tools

lachmanfrantisek:
- [x] tito (generating patches in source-git)
- [x] rdopkg (generating patches from the side-branch)
- [ ] the kernel work
- [ ] ~fedpkg import~
- [x] rpkg
- [x] rpkg-util https://pagure.io/rpkg-util
- [ ] rebase-helper
- [x] dlrn: https://github.com/softwarefactory-project/DLRN
- [ ] obs (open build service)

----

Anything else?

PR #45, opened 2020-05-19
Investigate how to split the Celery task in packit-service

lbarcziova:
This contains:
- summary of the discussion related to task splitting on arch day 
-  ideas/suggestions after going through the current codebase how to split the task and how to achieve it in smaller steps

NOTE: code snippets are just demonstrative

### DIFF
--- a/task-worklow-refactor/README.md
+++ b/task-worklow-refactor/README.md
@@ -6,13 +6,13 @@ will be the possibility to use more queues (for faster/slower tasks)
  and to differentiate tasks easily (Sentry).
 
 
-## Workflow
+## Workflow 1
  - get webhook payload or message from message bus 
  - add aditional info which can make determining the event type easier
  (GH webhooks - X-GitHub-Event)
  - send it to Celery as `task.process_message` - this task will take care about:
-    - parsing of the event into object of class Event (this class provides `project`, `package_config` properties,
-     which enables us to make the checks)
+    - parsing of the event into object of class Event (this class provides `project`, 
+    `package_config` properties, which enables us to make the checks)
     - private repository check
     - getting the handlers which handle the event
     - whitelist check
@@ -20,10 +20,11 @@ will be the possibility to use more queues (for faster/slower tasks)
     - (running `pre-check` here to filter out some events?)
     - sending specific task to Celery
   - handle specific tasks, which directly run the handlers
+ 
     
-#### How to pass the information needed by handlers
- Handlers are currently using the information from event (with `project` and `package_config` properties, specific for each event),
-  service config and job config object:
+#### How to pass the information needed by handlers (already discussed)
+ Handlers are currently using the information from event (with `project` and 
+ `package_config` properties, specific for each event), service config and job config object:
  ```python
 class JobHandler(Handler):
     def __init__(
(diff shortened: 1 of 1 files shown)


task-worklow-refactor/README.md
```diff
@@ -0,0 +1,94 @@
...
+ - add aditional info which can make determining the event type easier
+ (GH webhooks - X-GitHub-Event)
+ - send it to Celery as `task.process_message` - this task will take care about:
+    - parsing of the event into object of class Event (this class provides `project`, `package_config` properties,
+     which enables us to make the checks)
+    - private repository check
+    - getting the handlers which handle the event
+    - whitelist check
+    - possibly creating records in DB (explained below)
+    - (running `pre-check` here to filter out some events?)
```
lachmanfrantisek, line 20: Good point. We have `pre-check` both for Events and Handlers. It makes sense to run them before creating the new task.

task-worklow-refactor/README.md
```diff
@@ -0,0 +1,94 @@
...
+    - sending specific task to Celery
+  - handle specific tasks, which directly run the handlers
+    
+#### How to pass the information needed by handlers
+ Handlers are currently using the information from event (with `project` and `package_config` properties, specific for each event),
+  service config and job config object:
+ ```python
+class JobHandler(Handler):
+    def __init__(
+        self, config: ServiceConfig, job_config: Optional[JobConfig], event: Event
```
lachmanfrantisek, line 30: To make it a bit easier, we can move `config` and `job_config` to the `event`.

task-worklow-refactor/README.md
```diff
@@ -0,0 +1,94 @@
...
+ Possible solutions, which can be somehow combined:
+ 1. serialize the info about objects and pass it into `send_task`
+     - this would need serializing and then again deserializing
+ 2. save the info about objects in DB and pass IDs of models into `send_task`
+     - what models does make sense to have? possibilities:
+         - project
+         - package config 
+         - service config
+         - job config 
+         - event
```
lachmanfrantisek, line 42: Most of them can be calculated from the attributes of the event. => We can save only those attributes, that are needed for initialization of the event class. (And the information about event type.)

Saving and recreating the event instances is probably not the cleanest way how to do it.

jpopelka:
When looking at the `Which tasks do we want to have?` I had an idea that we could perhaps transform it into one matrix/table where rows would be tasks and columns would be jobs, handlers and events.
But it'd probably be too big and thus not that clear after all.

task-worklow-refactor/README.md
lachmanfrantisek: - `config`, `package_config` and `job_config` can be easily serialize
- `project` can be saved to the database and instantiate once again from the `config`. (We can forward database id.)
- `event` will cause more problems, it contains some metadata about the event itself (used in handlers) and we are using its type during the workflow to differentiate the responses.

task-worklow-refactor/README.md
```diff
@@ -0,0 +1,94 @@
...
+     
+     - each subclass of class Event stores different set of info -> does it make sense to create model for each? 
+     - these could be reused since in `task_results` table we store the event dict
+ 3. pass just the arguments which are required by the specific handler
```
lachmanfrantisek, line 46: This one will probably be the easiest one.

task-worklow-refactor/README.md
sakalosj: config and job_config should not be part of Event

csomh: Agree with @sakalosj here. Packing up objects might help with writing less code, and making it more easy to remember the arguments to be passed to calls, but it makes it more difficult to read and understand code. And we spend much more time reading code, than writing it, so it's more important to optimise for the reading part.

If something is called "event", it should have job configuration or service configuration stored in it.

task-worklow-refactor/README.md
csomh: This would go with nr 1, right?

I feel that going with the DB solution would lead to a more fragile solution, as the DB would become a single point of failure: no tasks—not even the ones not depending on calls to DB—could be created or processed.

Also, the serialisation problem seems to be a much easier one to solve than coming up with the right DB models.

sakalosj: from my point of view #3 is best option - there will be very clean boundaries between objects.

task-worklow-refactor/README.md
lachmanfrantisek: I confused that with `package_config`, my bad...

task-worklow-refactor/README.md
lachmanfrantisek: Yes, this will be much cleaner, but we need to forward all the information that is provided by `Event` instances. (And make some nice abstraction on this to not have a high number of arguments.)

And parts of the handlers' code is, unfortunately, event-dependent. These I can remember now:

- how to get a package config (easily forwardable)
- which project we want to work with (forwardable with some work)
- reporting (e.g. comments are nor possible for branch-push/release)
  - some abstraction should be done here (e.g. instead of event type, have `is_commenting`)

lachmanfrantisek: We can maybe use also the trigger models from the database.

sakalosj: I know exactly all this dependencies and by this i am suggesting to break them. It was main source of issues in CentOS/Pagure implementation.

e.g.
[CoprBuildJobHelper](https://github.com/packit-service/packit-service/blob/9e455f253f77ecda028473c49aa67b04c9927bff/packit_service/worker/build/copr_build.py#L52) 
* why it has Event dependency? In real world you can use copr api without Event, why dont do it that way? It will follow [OOP desgin principles](https://deviq.com/solid/)

* Also it contains "unrelated" methods (e.g. [self.report_status_to_all_for_chroot](https://github.com/packit-service/packit-service/blob/9e455f253f77ecda028473c49aa67b04c9927bff/packit_service/worker/build/copr_build.py#L159-L164)) which requires unrelated properties (e.g. [project](https://github.com/packit-service/packit-service/blob/9e455f253f77ecda028473c49aa67b04c9927bff/packit_service/worker/build/build_helper.py#L236-L242) ) , it is making from CoprBuildJobHelper some kind of [semi-God object](https://en.wikipedia.org/wiki/God_object)

lachmanfrantisek: @sakalosj I answered most of it already here and during the arch-day.

Some notes:
- We already agreed that we want to rename and clear `BuildHelper` classes.
- `BuildHelpers` do not run build itself, it actually calls the methods on `packit.PackitAPI`. (So its purpose is to trigger build and report the status in the meantime. So, It's not so God-like.)
- Don't forget that this task was meant to be the first step -- we can't rewrite the whole service here...;) It should be doable and reviewable.

---
Off-topic: When you regularly cites the OOP principles, it feels like you consider us/me very stupid. I know them.

sakalosj: @lachmanfrantisek 
sorry, didn't want to hurt anyone. It wasn't meant to attack, just to discuss my opinions, which can be wrong and this is one of platform to discuss.

Per my understanding this is research, and i think it should mentioned here. It doesn't mean it  have to be implemented all at once.

I used term "some kind of semi-God object", to express difference, that that there is something more that should be, but it is not  pure "God object".

I understand it like, [reporting](https://github.com/packit-service/packit-service/blob/9e455f253f77ecda028473c49aa67b04c9927bff/packit_service/worker/build/copr_build.py#L159-L164)) is acting upon one of git projects, therefore status reporter should be part of (or directly communicating with) class representing this git project. If we move it to BuildHelper we require also to pass project related informations. Currently it is done by adding project and config info to Event. Therefore I called it semi-God.

I think there is few places to improve, and because OOP principles is well known terminology, to cite them is easiest way to cover ideas i want to share.

TomasTomecek:
When I read the proposals, all I can see is to break class SteveJobs down to smaller units, which I completely agree with, since it's composed of multiple methods which are not too related to each other.

csomh:
I think I would prefer *Workflow 1*, just b/c all the small things (whitelist check, private repository check, and so on) are done in one place: `process_message`. This way we could avoid queuing multiple Celery tasks which then would not actually do the thing they supposed to do, b/c they would all fail on the same reason (a whitelist check, for example).

One more thing that would be cool to document here, is how the transition to this new design could be done in small steps. What would be the first change that would need to be done? It would be great to be able to make this transition with a series of small PRs, after each of which Packit Service is still functional.

lbarcziova:
@csomh I agree, for me, the 1. workflow is also more reasonable. I have added some notes about how can this workflow be achieved in smaller steps.

csomh:
> @csomh I agree, for me, the 1. workflow is also more reasonable. I have added some notes about how can this workflow be achieved in smaller steps (does it make sense to create 3 cards from the 3 steps?)

Yes :+1:

PR #46, opened 2020-05-20
Research on monitoring

csomh:
This will work in tandem with #43.


csomh:
@TomasTomecek anything else we should capture here? Is it okay to merge?

PR #52, opened 2020-06-29
ps_deployment_research

sakalosj:
Packit-service deployment improvement research.


ps_deployment/README.md
```diff
@@ -0,0 +1,119 @@
...
+		* stg every hour
+
+## Research
+### Installation source
+		
+	* github
+		* pros:
+			* current changes in other projects are always in place - useful especially for stg branch
+		* cons:
+			* (please add if you are aware of any)
```
lachmanfrantisek, line 36: I am thinking only about this:

```suggestion
			* Handling of dependencies is harder (we need to mix `rpm` and `PyPI` versions).
```

ps_deployment/README.md
```diff
@@ -0,0 +1,119 @@
...
+	
+	* rpm:
```
lachmanfrantisek, line 38: And the opposite of the previous one:

```suggestion
	* rpm:
		* pros:
			* All dependencies are installed similarly and automatically via `rpm`.
```

ps_deployment/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+	* Image Stream - current configuration
+	* Custom Build - need more details
+
+## Conclusion
+
+	* Source-to-image (S2I) - no performance improvements, bigger image
+	* RPM vs github:
+		* from performance perspective basically same results 
+		* github repos should not have issues with OS version - will make deployment easier
```
lachmanfrantisek, line 108: We had those problems in the past...

And we will have different dependencies for CLI and for service.

Alternative solution mixing those two:
- Installing from copr-repo that is built from `master`/`stable` branch. (I set up that last week.)
- :heavy_plus_sign: rpm versions for everything
- :heavy_plus_sign: no waiting for Fedora process 
- :heavy_minus_sign: We can't react to new commit to branch -- the build is not finished in that time. (Yes, this can be a problem.)

ps_deployment/README.md
```diff
@@ -0,0 +1,121 @@
...
+	* splitting image to base + final, will improve build times 
+	
+	AI:
+	
+
+## TODO
+	* trigger OC pull automatically after image is build
+	* check dockerhub alternatives
+	* using cache / forcing rebuild (e.g. pip install may install new version, yaml file is updated - force ansible to re-initiate ansible command)
+	* if base image is updated - re-trigger - how to implement this in some pipeline
```
jpopelka, line 120: For example, Docker Hub has a 'Repository Links' feature, which if set to 'Enable for Base Image' should trigger a build whenever a base image is updated. But it seems to work only for `Dockerfile`, but not for example for `Dockerfile.worker` (like [here](https://hub.docker.com/repository/docker/usercont/packit-service-worker/builds/edit))

csomh: Maybe we could live without having anything like this ATM: images are rebuilt fairly often, so they would pick up updates in the base image relatively fast.

ps_deployment/README.md
```diff
@@ -0,0 +1,121 @@
...
+	* [s2i strategy options](https://docs.openshift.com/container-platform/3.11/dev_guide/builds/build_strategies.html#source-to-image-strategy-options)
+	* [s2i tool blog](https://www.openshift.com/blog/create-s2i-builder-image)
+
+	* Image Stream - current configuration
+	* Custom Build - need more details
+
+## Conclusion
+
+	* Source-to-image (S2I) - no performance improvements, bigger image
+	* RPM vs github:
```
csomh, line 108: This section is relevant for `[python]-packit` and `python-ogr`, right?

Could/should we treat these two packages separately?

When installing from source with `pip` we should pay attention to how we pull in dependencies from PyPI, as those packages are not signed. Without enabling extra checks, this can become a fairly easy attack vector.

ps_deployment/README.md
```diff
@@ -0,0 +1,121 @@
...
+		* from performance perspective basically same results 
+		* github repos should not have issues with OS version - will make deployment easier
+	* splitting image to base + final, will improve build times 
+	
+	AI:
```
csomh, line 113: I would start with:

* creating a base image with Ansible to be shared by the service, worker and listener images (and any other image we have that needs it).
* breaking up the dependency between the `packit` and the worker images, and maybe even removing the `packit` image. I would say to install `packit` from source in the workers, from the `master` branch when building for stage and from `stable` when building for prod.
* Check the results and see if further layering or refinement of the build process is required.

What do you think?

ps_deployment/README.md
sakalosj: thanks will look into it

ps_deployment/README.md
jpopelka: I agree completely, with one note:

>     * would say to install `packit` from source in the workers, from the `master` branch when building for stage and from `stable` when building for prod.

Or, as Franta pointed out, we [now](https://github.com/packit-service/packit/pull/879) should have [packit-master](https://copr.fedorainfracloud.org/coprs/packit/packit-master) & [packit-releases](https://copr.fedorainfracloud.org/coprs/packit/packit-releases) (doesn't exist yet, since no builds have been done yet) repos in Copr, which should contain the packit & ogr rpms for their respective `master` & `stable` (in `ogr` case it means the PyPI release) branches.

The benefit I see in installing RPMs instead of from git is that RPMs will take care of all the dependencies so we don't have to have them separately in some playbook.

sakalosj: creating a base image with Ansible to be shared by the service, worker and listener images (and any other image we have that needs it).

https://projects.engineering.redhat.com/browse/PACKIT-443

    breaking up the dependency between the packit and the worker images, and maybe even removing the packit image. I would say to install packit from source in the workers, from the master branch when building for stage and from stable when building for prod.

https://projects.engineering.redhat.com/browse/PACKIT-444

    Check the results and see if further layering or refinement of the build process is required.
i think, yes - it will be essential, but will support it with data once card above are implemented

ps_deployment/README.md
sakalosj: is here option to install 3rd party via rpm (where applicable - some are using pip install as workaround) and install our projects via pip with  --no-deps?

ps_deployment/README.md
sakalosj: please check https://github.com/packit-service/research/pull/52#discussion_r448282810 - there is discussed same topic

TomasTomecek:
I'm not gonna inline comment so this wouldn't blow completely, I'd say you guys are doing solid job there.

My overall thought here is to realize why we are doing this - to improve velocity on PRs, deployment and while doing local testing - getting builds and test results faster. I'd love if that thought was the main driver here. If the outcome of the research are cards which change the image layout/automation with zero difference in the velocity, I'm gonna NACK such result very hard. I feel like Hunor is getting this (re the ansible base image idea) so overall all's good - this is not a rant to be completely clear. My $.02.

ps_deployment/README.md
csomh: > The benefit I see in installing RPMs instead of from git is that RPMs will take care of all the dependencies so we don't have to have them separately in some playbook.

Yes, this is a great benefit to have. On the other hand I don't see this approach to work when someone would need to build  the Packit Service images locally with some custom modifications in `packit` and/or `ogr`. Or am I missing something.

With `pip install` the above would be possible, we just need to expose the source for these packages as build arguments, defaulting to the GitHub repos, `ADD` these sources into the image, and `pip install` them.

I agree that maintaining the rpm dependencies in the playbooks is far from ideal. Could we maybe get a list of dependencies from `setup.cfg` and then install them with `dnf install 'python3dist(<name>)'`?

jpopelka: > I don't see this approach to work when someone would need to build the Packit Service images locally with some custom modifications

agree, let's scratch my idea

> Could we maybe get a list of dependencies from `setup.cfg` and then install them with `dnf install 'python3dist(<name>)'`?

Not any straightforward I know of. Plus there are [exceptions (bodhi-client and koji required by packit and not in setup.cfg)](https://github.com/packit-service/packit/blob/master/setup.cfg#L59)

ps_deployment/README.md
lachmanfrantisek: As discussed today:
- We have some dependencies both on PyPI and in Fedora and some Fedora only. => It would be nice to use only one source, the Fedora one.
  - Also for other reasons like signing as Hunor said or quality (packages in Fedora has fewer problems with installations than the PyPI ones since it's prepared for our environment and can use Fedora dependencies).
  - Another positive point is matching with packit-cli. We already had some problems with rebase-helper when Fedora and PyPI version differed.
- We should avoid adding a new place where we need to manage dependencies.
- We can:
  - Make a local build of Packit (via packit-cli or manually).
  - Try to install only dependencies using something like [this](https://github.com/packit-service/packit/blob/master/files/tasks/build-rpm-deps.yaml) and use `pip` with `--no-deps`.
    - note: Are `builddeps` enough?

Question for others: which version of ogr we want to use? Is `master` (the current state) ok?

jpopelka: >     * Are `builddeps` enough?

No, they're not. But I have one heretic idea. What about adding those missing
```
Requires:       fedpkg
Requires:       rpmdevtools
Requires:       python3-bodhi-client
```
as
```
# These are technically not needed for the build,
# but we (packit team) need to have them here to be able to install all dependencies with
# dnf -y builddep packit.spec
BuildRequires:       fedpkg
BuildRequires:       rpmdevtools
BuildRequires:       python3-bodhi-client
```
as well into the [spec](https://github.com/packit-service/packit/blob/master/packit.spec) as well.
I don't think this (having more build requires than needed) is against [packaging guidelines](https://docs.fedoraproject.org/en-US/packaging-guidelines/#buildrequires).

lachmanfrantisek: @jpopelka There is no `requires` version of that `dnf` command, is it?

I am not against your change otherwise. (And since we want to run real tests in the `test` step, we need to have those dependencies in all cases.)

csomh: It might create a lot of confusion in the future if we start mixing build requirements with run-time requirements. These to categories overlapping is more accidental than intentional, so relying on it could cause unnecessary fragility.

Another approach we could take:

With the help of `rpmspec` one can get the run-time requirements from the spec file with

```
rpmspec -q --requires packit.spec | grep -v 'packit'
```

The grep is to get rid of `python3-packit` as a dependency, as that one is provided by us anyways.

This way we could get the run-time requirements that are specified only in the spec file.

We would still miss run-time requirements from `setup.cfg`. These could be obtained by running `python setup.py egg-info`, which would give us `packitos.egg-infor/requires.txt`. From this file we could install each run-time dependency as `dnf install python3dist(<dep>)`, of course after dropping the `[testing]` section. This mechanism would be similar to what Fedora does when [auto-generates Python dependencies](https://docs.fedoraproject.org/en-US/packaging-guidelines/Python/#_automatically_generated_dependencies).

lachmanfrantisek: Looks like local build will be the easiest. (We can run `rpmbuild` directly if we don't want to install packit.)

It will avoid duplicating dependencies definition, parsing and other magic which can be confusing and can have errors.

sakalosj:
> I thought that we'd already merged this.

I forgot about it ... thanks for approval

PR #53, opened 2020-07-16
Fix regex for searching conditional patching

dhodovsk:


centos_playground/pkg_survey/survey.py
```diff
@@ -72,6 +74,18 @@ def do_mock_build(self):
         shutil.copyfile('/var/lib/mock/centos-stream-x86_64/result/build.log', err_log_file)
         self.pkg_res['error'] = f'mock build failed. More info in: {err_log_file}'
 
+    @staticmethod
+    def get_conditional_info(spec_cont):
+        conditions = re.findall(r'\n%if.*?\n%endif', spec_cont, re.DOTALL)
+        result = []
+        for con in conditions:
+            if '\n%patch' in con:
+                p = re.compile("\n%if (.*)\n")
```
lachmanfrantisek, line 83: This can be shared between all conditions or ideally all the packages...

centos_playground/pkg_survey/survey.py
```diff
@@ -72,6 +74,18 @@ def do_mock_build(self):
...
+                found = p.search(con)
+                if found:
+                    result.append(p.search(con).group(1))
```
lachmanfrantisek, line 86: You don't need to `search` twice, just use `found` instead:

```suggestion
                found = p.search(con)
                if found:
                    result.append(found.group(1))
```

PR #60, opened 2020-09-21
Add proposal on how to deploy changes in a test environment

csomh:


deploy-packit-pr/README.md
```diff
@@ -0,0 +1,120 @@
...
+## Where to build the images?
+
+* Docker Hub
+    - Slow; there can be only one build in our project, since we are using a
+      free account.
+    - It's not that obvious how to trigger on-demand builds from PRs.
+        - Actually, @sakalosj told that it's possible to trigger builds using the
+          Docker Hub API, but the method is not documented.
+* Quay
+    - Similar to Docker Hub.
```
TomasTomecek, line 67: since Quay is a service maintained by Red Hat, I'd love to know if we can switch to using it (and do the switch, e.g. in Q4)

deploy-packit-pr/README.md
```diff
@@ -0,0 +1,120 @@
...
+At least in the beginning we might not want to deploy every single PR, so we
+might need to have an explicit way for triggering this.
+
+**Minimal solution:** write a script taking some arguments, which each of us
+can run locally. For this each of us would need to have the credentials to the
+environment(s) the script needs to interact with.
+
+**One step further:** have some service/bot, listening to GitHub
+comments/label changes and running the deployment script. We could check with
+softwarefactory folks if we could get a new Zuul pipeline to handle this.
```
TomasTomecek, line 20: also github actions could be used

TomasTomecek: but the thing is that we need to share all our secrets to that environment

csomh: made a note about this.

deploy-packit-pr/README.md
csomh: Filed packit/packit-service#833 to look into this.

deploy-packit-pr/README.md
```diff
@@ -0,0 +1,122 @@
...
+        - Actually, @sakalosj told that it's possible to trigger builds using the
+          Docker Hub API, but the method is not documented.
+* Quay
+    - Similar to Docker Hub.
+* OpenShift Online
+    - Costs us money.
+    - But we could use resource limits to have it under control.
+* PSI
+    - Behind the firewall.
+    - But builds could push images to a public registry.
```
csomh, line 75: What do you think @packit/the-packit-team? Which approach would work the best?

csomh (commented):
@packit/the-packit-team Any comments on this? Is this good enough to start creating stories based on it and get to work?

deploy-packit-pr/README.md
jpopelka: I'd start with the one which is easiest to implement, but I'm not sure which one that is ;-)
```python
If @sakalosj is right (that it's possible to trigger a build via API):
  then it's probably the Docker Hub, even it's also the slowest option
else:
  if money is no problem:
    then OpenShift Online
  else:
    PSI - but how would we let the service/worker in PSI know, via some public broker? Or would it be checking the repos periodically?
```
If not, then

jpopelka (approved):
> Is this good enough to start creating stories

yes, we've discussed this enough, I'd say

deploy-packit-pr/README.md
csomh: Thanks for the feedback @jpopelka!

Tbh, I'm a little bit concerned to go with the Docker Hub API approach. As @sakalosj told us, it's not documented, and I see some risks due to this: we might need to spend time to figure out how it works and we won't have any guaranties that it stays the same over time.

If we trigger the deployments by a script, then building in PSI should not be a problem. Maybe we could use this state to better understand the resource requirements of building. Once we have a bot built around it, we can easily migrate the build configs to somewhere public (OpenShift Online). WDYT?

jpopelka: Ok, I forgot about the trigger-by-script workflow. Then PSI looks OK.

sakalosj: I am for PSI, but in future, if we want to move some automation to zuul, it will be required to have it publicly accessible

PR #62, opened 2020-10-06
pre-commit

lachmanfrantisek:
- I'm really used to use pre-commit so I've added the config for that and reformat the files.
  - The content of the config is the same as we have for packit.
  - mypy is ignored for now since I could not ignore the `Duplicate module named` error (and don't want to change the setup).


lachmanfrantisek:
As you can see, we have a really huge amount of code here..;)

What about excluding the onboarding script(s) to a separate project. It's getting bigger and it's no longer A script. I think it deserves a separate home. WDYT? (Espetially @dhodovsk as an author?)

jpopelka (approved):
> mypy is ignored for now

I'm wondering whether we need black, flake8 and mypy at all in this mostly documentation repo.
I'd consider all code here a proof of concept, hence not needing any validation.

> What about excluding the onboarding script(s) to a separate project.

+1

lachmanfrantisek:
> I'm wondering whether we need black, flake8 and mypy at all in this mostly documentation repo.

My reason was the onboarding script which receives some care continuously.

Can I include the black reformat changes for now and remove the black, mypy and flake8 from the pre-commit once we exclude the onboarding script?

jpopelka:
> Can I include the black reformat changes for now and remove the black, mypy and flake8 from the pre-commit once we exclude the onboarding script?

sure

PR #69, opened 2020-12-01
Update of source-git from upstream

lachmanfrantisek:
Signed-off-by: Frantisek Lachman <flachman@redhat.com>

The rendered version can be seen [here](https://github.com/lachmanfrantisek/research/tree/upstream-to-source-git-update/source-git).


source-git/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+### How do we get known about a need for a rebase?
+
+Various ideas about the trigger for rebasing follow. Combination of multiple approaches is possible.
+
+#### 1. Manual update
+
+- User decides when he wants to do the rebase.
+- Implementation:
+  - Using a packit CLI.
```
TomasTomecek, line 66: this is probably the best for the new workflow with upstream git history

source-git/README.md
```diff
@@ -0,0 +1,89 @@
...
+  - New issue in source-git with a title like `rebase from upstream`.
+
+#### 2. Scheduled update
+
+- No need to listen to any service.
+- We can potentially use Upstream Release Monitoring.
+- Like the current updates.
```
TomasTomecek, line 73: this would be interesting to do in rawhide for the "automaintainer" use case by opening PRs against source-git with the new updates

source-git/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+#### 3. Listen for upstream events
+
+- We don't have access to upstream projects.
+- Upstream will either need to install our GitHub application or be enabled on github2fedmsg.
```
TomasTomecek, line 78: this seems like no-go right now, since many projects are not even on github, gitlab, pagure.io

source-git/README.md
```diff
@@ -0,0 +1,89 @@
...
+- For each new upstream release, we will regenerate source-git and force-push.
+
+<img src="./img/force-push-upstream-release.svg" alt="force-push-upstream-release" height="200"/>
+
+#### 3. Revert & Regenerate
+
+- We can use same solution we discussed [updates from distgit](../dist-git-to-src-git/updates.md).
+
+<img src="./img/revert-and-regenerate-upstream-release.svg" alt="revert-and-regenerate-upstream-release" height="200"/>
+
```
TomasTomecek, line 39: how about doing the new update in a dedicated branch and then git-merging it?

it also seems like things will be different when the source-git repo has an upstream history and when it doesn't

source-git/README.md
```diff
@@ -0,0 +1,89 @@
...
+#### 2. Force-push
+
+- For each new upstream release, we will regenerate source-git and force-push.
+
+<img src="./img/force-push-upstream-release.svg" alt="force-push-upstream-release" height="200"/>
+
+#### 3. Revert & Regenerate
+
+- We can use same solution we discussed [updates from distgit](../dist-git-to-src-git/updates.md).
```
TomasTomecek, line 36: I personally like this the most for the current source-git repos

source-git/README.md
```diff
@@ -0,0 +1,89 @@
...
+- [ ] The changed source-git can be converted to dist-git.
+  - Via PR or a new Bugzilla.
+  - (This text does not cover this.)
+
+What changes can occur?
+
+- Change in upstream means always change in the source-archive in distgit.
+- Change in source-git can mean:
+  - Change in distribution files => can be converted directly.
+  - Change in the files => new patch needs to be added.
```
TomasTomecek, line 17:
```suggestion
  - Change in the source code => new patch needs to be added.
```

source-git/README.md
lachmanfrantisek: > how about doing the new update in a dedicated branch and then git-merging it?

Will this work?

* Can we mix source-git commits with the upstream ones?
* We need to have full git-history for the generation. (Will not be preserved by merge.)

TomasTomecek: for 9 it will definitely work

for 8 it seems recreate is the way to go - I wonder if we can move some of the code from d2s to packit to do as many manual steps as possible

would you have time to update this research wrt difference b/w 8 and 9 source-git repos?

lachmanfrantisek: Sorry, I missed your last comment and not sure if I get it correctly. The following scenario is not supported by packit, is it? (Or a different question, do we want to support it?)

```
EXAMPLE SOURCE-GIT HISTORY:

* downstream-patch commit
* upstream commit
* upstream commit
* downstream-patch commit
* downstream packaging files
* upstream commit
:
```

lachmanfrantisek:
@TomasTomecek PTAL

- The part about rebasing vs regeneration has been added.
- An example of a rebase for chrony.
- Part about Python source-git.
- Some conclusion (might be enhanced after the review).

source-git/README.md
TomasTomecek: depends what you mean by supported, upstream_ref is the king here - whatever is behind, packit doesn't care, whatever is ahead, it will be processed

so you're right that git-merge is not the way to go, we either need to force-push (recreate) or create branches per rebase

source-git/README.md
```diff
@@ -0,0 +1,180 @@
...
+<img src="./img/sync-upstream-release+revert-and-regenerate.svg" alt="sync-upstream-release+revert-and-regenerate" height="200"/>
+
+### To rebase or not to rebase?
+
+It is a question if we want to do rebase or do the part of the source-git generation again.
+How this can differ?
+
+- Our generation algorithm can be improved.
+- The default packit config can be improved.
+- The non primary sources can change. (We add them in a form of commit on top of the upstream history.)
```
TomasTomecek, line 65: this is totally a concern for multiple projects - the rebase part needs to probably rerun %prep and then rebase or `reset --hard`

source-git/README.md
```diff
@@ -0,0 +1,180 @@
...
+Various ideas about the trigger for rebasing follow. Combination of multiple approaches is possible.
+
+#### 1 Manual update
+
+- User decides when he wants to do the rebase.
+- Implementation:
+  - Using a packit CLI.
+- Possible conflicts can be solved on a git level.
+
+#### 2 Automation
```
TomasTomecek, line 78: We should start with doing this manually, define the process/workflow, write code which makes this easy and then over time start thinking how to do some mega-automation.

source-git/README.md
```diff
@@ -0,0 +1,180 @@
...
+- Fedora patches stored in https://github.com/fedora-python/cpython
+- In the repo, there are `fedora-X.Y` branches for python version`X.Y`.
+- This branch is based on the upstream history (build on top of the upstream tag).
+- On top of the upstream history, there are patches in form of git commits.
+- There is a naming scheme for the patch commits.
+- The git rebase is used for updating the branches. (There is a tag left for history purpose.)
+- `git cherry-pick` used for rebasing downstream patch commits from a different branch.
+- Fix-up commits are created when patch needs to be edited.
+  (Interactive rebase is used when generating the dist-git patches.)
+- Removal is done as a commit revert.
```
TomasTomecek, line 131: NICE, this is literally what we need :)

TomasTomecek (approved):
Very nice, thank you for being so detailed, this is gonna make things so much easier for the next work.

source-git/README.md
csomh: @TomasTomecek I'm a little bit confused now: shouldn't this be only about 9, and that is only about source-git repos which connect to the upstream history? What am I missing?

lachmanfrantisek: > supported

Packit can work with such repo => srpm/rpm can be created & bugzilla/PR to distgit can be generated

lachmanfrantisek:
I am going to merge this but we can continue the discussion. Thanks for the review!

PR #74, opened 2021-02-08
Monitoring of Openshift objects

lbarcziova:


monitoring/README.md
```diff
@@ -46,6 +46,85 @@ metrics from the workers and provide the `/metrics` endpoint for Prometheus.
    Pushgateway] to be set up and configured for scraping.
 3. Monitor some resources.
 
+### How to monitor the Openshift objects?
+
+1. [`kube-state-metrics`](https://github.com/kubernetes/kube-state-metrics)
+
+   - converts Kubernetes objects to metrics consumable by Prometheus
+   - not focused on the health of the individual Kubernetes components, but rather on the health of the various objects inside, such as deployments, nodes and pods
```
TomasTomecek, line 54: does this imply it focuses on application pods, services, deployments, controllers etc. instead of the functional kubernetes parts such as nodes, container runtime, master api, container registry...

monitoring/README.md
```diff
@@ -46,6 +46,85 @@ metrics from the workers and provide the `/metrics` endpoint for Prometheus.
...
+   - metrics are exported on the HTTP endpoint `/metrics` on the listening port, designed to be consumed either by
+     Prometheus itself or by a scraper that is compatible with scraping a Prometheus client endpoint
+   - deleted objects are no longer visible on the `/metrics` endpoint
+   - examples of exposed metrics (more [here](https://github.com/kubernetes/kube-state-metrics/tree/master/docs)):
+     - kube_pod_container_resource_requests
+     - kube_pod_container_status_restarts_total
+     - kube_pod_status_phase
+   - in limited-privileges environment where you don't have cluster-reader role:
```
TomasTomecek, line 62: I hope they have clear docs how to set it up because roles in k8s are always a nightmare to figure out properly

monitoring/README.md
```diff
@@ -46,6 +46,85 @@ metrics from the workers and provide the `/metrics` endpoint for Prometheus.
...
+     - create a serviceaccount
+     - give it view privileges on specific namespaces (using roleBinding)
+     - specify a set of namespaces (using the --namespaces option) and
+       a set of kubernetes objects (using the --resources) that your serviceaccount
+       has access to in the kube-state-metrics deployment configuration
+   - example deployment configuration files in their Github repo [here](https://github.com/kubernetes/kube-state-metrics/blob/master/examples)
+   - can be configured in the args section of the deployment configuration using
+     [CLI args](<(https://github.com/kubernetes/kube-state-metrics/blob/master/docs/cli-arguments.md#command-line-arguments)>)
+
+2. [Node exporter](https://github.com/prometheus/node_exporter)
```
TomasTomecek, line 72: this could potentially help us with the random filesystem problems we are experiencing

TomasTomecek (commented):
very nice write-up!

monitoring/README.md
lbarcziova: yes, right

monitoring/README.md
csomh: I think this is not an option in OpenShift Online, as we don't have access to the nodes.

csomh (approved):
Thank you @lbarcziova for doing the research!

In OpenShift Online we have probably kube-state-metrics, kubernetes_sd_config and maybe cAdviser as an option. Although there might be issues with the kubernetes version being very old (1.11).

I would say this can be merged and let's discuss these findings on the next arch meeting.

monitoring/README.md
lbarcziova: yes, in some options there will be problem with access in current Openshift Online, but I wanted to cover more options also for future, since we were thinking about migrating from Online

PR #77, opened 2021-02-16
Copr custom source method

lbarcziova:
Example: https://copr.fedorainfracloud.org/coprs/lbarczio/ogr-test-custom/build/1967579/

### DIFF
--- a/building-srpm-in-copr/README.md
+++ b/building-srpm-in-copr/README.md
@@ -1,0 +1,92 @@
+Copr provides an option to use their [custom source method](https://docs.pagure.org/copr.copr/custom_source_method.html)
+to create sources for SRPM.
+
+## How to use the custom source method to create sources
+
+- required script in any scripting language and chroot where the script is executed (defaults to `fedora-latest-x86_64`)
+- output of the script: specfile, optionally any other file needed to successfully build a source RPM from that spec file (tarballs, patches)
+- script is executed under non-privileged user -> packages needed for running script should be specified as a list of (srpm)build-dependencies
+- optional parameters:
+  - builddeps: list of packages
+  - resultdir: dir with output files (defaults to current working directory)
+- cannot be run separately without creating a copr build
+  (we cannot use it for koji builds)
+- in `python-copr` we would call this method (currently we use `create_from_file`):
+
+```python
+    def create_from_custom(self, ownername, projectname, script, script_chroot=None,
+                           script_builddeps=None, script_resultdir=None, buildopts=None,
+                           project_dirname=None):
+        """
+        Create a build from custom script.
+
+        :param str ownername:
+        :param str projectname:
+        :param script: script to execute to generate sources
+        :param script_chroot: [optional] what chroot to use to generate
+            sources (defaults to fedora-latest-x86_64)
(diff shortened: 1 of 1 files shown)


building-srpm-in-copr/README.md
```diff
@@ -0,0 +1,92 @@
...
+         used also here, e.g. `ActionName.create_archive`
+    3. `create_srpm(srpm_path=output_file, srpm_dir=srpm_dir)`
+       - run `rpmbuild` command
+- we could create a CLI command in Packit which would clone the repo in the version we want (particular branch, commit sha), run 1. and 2. step from current process,
+  move the needed files into `resultdir`
+- we need to get the required info into the script:
+  - git ref
+  - project url for cloning the repo and additional arguments to clone the version we want
+- the script could look like this, `packit` would be the build dependency
+  (+ packages needed for running the user-defined actions?):
```
csomh, line 57: With this we could solve the build-dependency issues, and let users define them.

building-srpm-in-copr/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+```bash
+#!bin/sh
+
+packit prepare-sources --project_url={PROJECT_URL} --upstream-ref={UPSTREAM_REF}
```
csomh, line 62: `PROJECT_URL` and `UPSTREAM_REF` could be arguments, instead of options, but I guess that's a matter of preference :)

csomh (approved):
Thanks for this very hands-on write-up! :rocket: 

I think having this approach could help us offload some of the work currently Packit Service does to Copr.

This would be nice as it would help us get rid of an infrastructure dependency (sandcastle pods) and with less bytes being coppied around, it could even help speeding up things a little (although some of that would be taken away by dependencies being installed before every SRPM build).

I'll add this as a topic for the next arch to discuss what to do next.

building-srpm-in-copr/README.md
```diff
@@ -0,0 +1,92 @@
+Copr provides an option to use their [custom source method](https://docs.pagure.org/copr.copr/custom_source_method.html)
+to create sources for SRPM.
+
+## How to use the custom source method to create sources
+
+- required script in any scripting language and chroot where the script is executed (defaults to `fedora-latest-x86_64`)
+- output of the script: specfile, optionally any other file needed to successfully build a source RPM from that spec file (tarballs, patches)
+- script is executed under non-privileged user -> packages needed for running script should be specified as a list of (srpm)build-dependencies
```
TomasTomecek, line 8: should the dependencies be specified in the spec file?

TomasTomecek: okay, script_builddeps is the answer to my question

building-srpm-in-copr/README.md
```diff
@@ -0,0 +1,92 @@
...
+- optional parameters:
+  - builddeps: list of packages
+  - resultdir: dir with output files (defaults to current working directory)
+- cannot be run separately without creating a copr build
+  (we cannot use it for koji builds)
```
TomasTomecek, line 13: We could request this functionality be implemented in copr - though hard to say if the copr team would find this useful and how long it would take to implement

jpopelka (approved):

PR #80, opened 2021-03-16
Add notes and research on error budgets

csomh:

### DIFF
--- a/error-budgets/README.md
+++ b/error-budgets/README.md
@@ -1,0 +1,319 @@
+# Error Budgets
+
+## Next steps for Packit
+
+1. Identify stakeholders who can help us to define our SLO
+
+   - Project which are the most frequent users of the service.
+   - Prominent users:
+     - [rhinstaller/anaconda](https://github.com/rhinstaller/anaconda)
+     - [nmstate/nmstate](https://github.com/nmstate/nmstate)
+     - [systemd/systemd](https://github.com/systemd/systemd)
+     - [oamg/convert2rhel](https://github.com/oamg/convert2rhel)
+     - [psss/tmt](https://github.com/psss/tmt)
+     - [avocado-framework/avocado](https://github.com/avocado-framework/avocado)
+     - [rpm-software-management/dnf](https://github.com/rpm-software-management/dnf)
+     - [fedora-modularity/libmodulemd](https://github.com/fedora-modularity/libmodulemd)
+     - [containerbuildsystem/atomic-reactor](https://github.com/containerbuildsystem/atomic-reactor)
+     - [OpenSCAP/oval-graph](https://github.com/OpenSCAP/oval-graph)
+     - [rpm-software-management/libdnf](https://github.com/rpm-software-management/libdnf)
+     - [containerbuildsystem/osbs-client](https://github.com/containerbuildsystem/osbs-client)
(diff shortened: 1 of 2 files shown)


error-budgets/README.md
```diff
@@ -0,0 +1,319 @@
...
+reproducible way.
+
+##### Forming an Error Budget
+
+Based on the service level objective (SLO).
+
+Defines a clear, objective metric that determines how unreliable a service is
+allowed to be within a quarter.
+
+1. Product Management defines the SLO, which sets and expectation for the
```
TomasTomecek, line 161:
```suggestion
1. Product Management defines the SLO, which sets an expectation for the
```

error-budgets/README.md
```diff
@@ -0,0 +1,319 @@
...
+   uptime.
+2. Monitoring system (neutral third party) measures the uptime.
+3. The previous two define the "budget" of how much "unreliability" is left in
+   the quarter.
+4. As long as there is error budget remaining, releases can be pushed.
+
+##### Benefits
+
+Discussions/decisions around release velocity can be driven by the error
+budget.
```
TomasTomecek, line 171: Wow, so we could actually deploy to prod multiple times a week :O

TomasTomecek (approved):
I already have questions for the SLIs and SLOs and will hold my horses for the discussion with stakeholders.

Since you pointed out so many high-volume projects - would a survey be an efficient way of gathering feedback?

error-budgets/README.md
```diff
@@ -0,0 +1,319 @@
...
+3. Figure out which of the expectations identified in the previous step can be
+   measured with an indicator. What are these indicators (SLIs)? What are the
+   SLOs?
+   - We might want to check whether the SLOs make sense to the stakeholders.
+4. Capture/produce the indicators (metrics/monitoring)
+5. Track indicators to know how we are doing against our targets (SLOs). This
+   means setting up monitoring dashboards and alerts, and making sure we have
+   the process in place to become aware and take action (if needed) in case
+   objectives are in danger to be missed.
+
```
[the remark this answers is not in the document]
csomh: Yes, exactly.

csomh:
> Since you pointed out so many high-volume projects - would a survey be an efficient way of gathering feedback?

Maybe, though I would probably find quick, one-on-one discussions or chats more useful. The reason for this is that surveys are not that good at finding the "unknown unknowns". A one-on-one discussion, being less orchestrated, can surface many of these. 

We don't necessarily need to interview every single of these projects. Getting the same answers over and over again or not learning anything new from subsequent interviews could be a good sign to know when to stop.

TomasTomecek:
> > Since you pointed out so many high-volume projects - would a survey be an efficient way of gathering feedback?
> 
> Maybe, though I would probably find quick, one-on-one discussions or chats more useful. The reason for this is that surveys are not that good at finding the "unknown unknowns". A one-on-one discussion, being less orchestrated, can surface many of these.
> 
> We don't necessarily need to interview every single of these projects. Getting the same answers over and over again or not learning anything new from subsequent interviews could be a good sign to know when to stop.

Agreed. My concern was that such conversations would be really time-consuming but you're right that we can stop with those any time and reach out to the rest via an email.

Thanks again for this write-up, seems to me this is exactly what we need at this point to understand how can we serve our users better and progress further with the project.

PR #82, opened 2021-03-23
Add research notes about prioritizing of the Celery tasks

lbarcziova:


jpopelka:
Thank you, that's very nice research.

I know the task was about researching "prioritization of tasks", I'm however not sure it would completely solve our problem.
It's however a very good start.

The problem, as I see it, is that we have tasks that take a long time to finish, for example, the task which creates an SRPM in sandcastle before submitting a copr build - I guess it's the `run_copr_build_handler`, which is in the `high-prio` column in your table. Such tasks can block the workers for a long time. Even if we had more workers serving the `high-prio` queue and only one worker serving the `low-prio` queue, the `low-prio` queue can actually be processed very quickly because all the tasks there take max few secs while some tasks in `high-prio` queue could wait in the queue for a long time if all the `high-prio` workers were processing the long-running `run_copr_build_handler` tasks.

What I think we need to investigate is not how to classify the tasks based on priority, but based on how long they usually take to process.
Then we can have `long-running` and `short-running` queues and have more workers serving the `long-running` queue and one worker serving the `short-running` queue. That way we can decrease the overall latency because all the short-running tasks would be processed almost immediately since no long-running task would be blocking them.

I'm assuming the `run_copr_build_handler` is the only long-running task we have. If we have more, then we would probably need to base on your research and have `long-running-high-prio` and `long-running-low-prio`.

From [Celery docs](https://docs.celeryproject.org/en/stable/userguide/optimizing.html#prefetch-limits):
"If you have a combination of long- and short-running tasks, the best option is to use two worker nodes that are configured separately, and route the tasks according to the run-time."

lbarcziova:
Thanks for the comment, these are really good points!

Yes, the current division in the PR is only based on which tasks are triggered by some user action, since the followup of my card should be packit/packit-service#836.

However, I will incorporate your points in this PR and also the division depending on the duration of the task and focus on that as well and we can talk about this on the architecture.

celery_tasks_prioritizing/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+`celery --app="${APP}" worker --loglevel=${LOGLEVEL} --concurrency=1 --prefetch-multiplier=1`
+
+## What do we want to achieve
+
+- process the tasks depending on their priority
+- higher priority: tasks which should provide faster feedback - tasks triggered by user actions,
+  after user pushes/creates PR/comments, he should see that p-s processes this action as soon as possible (sets flags, ...)
+- lower priority: reporting Copr build start, TF results, actions triggered by other events that don't require
+  fast feedback to user
```
TomasTomecek, line 16: one nit: could you please turn these into sublists so it's easier to read? e.g.
```
* higher priority:
  * tasks which should provide faster feedback - tasks triggered by user actions
  * ...
```

This is a very good list, I'd love to suggest a slight rewording to a few items (or maybe even interpretation)
* high prio:
  * actionable events to users such as results with logs, final events
  * informations with high value - i.e. first status report
* lower prio:
  * intermediate results
  * long running tasks

Thanks!

celery_tasks_prioritizing/README.md
```diff
@@ -0,0 +1,86 @@
...
+- there are 10 (0-9) priority levels
+- ```
+  app.conf.broker_transport_options = {
+    'queue_order_strategy': 'priority',
+    }
+  ```
+- creating `n` lists for each queue - by default 10 priority levels are consolidated to 4 levels
+  - e.g. queue named `celery` will be split into 4 queues:
+    ` ['celery0', 'celery3', 'celery6', 'celery9']`
+  - for more levels - `priority_steps` transport option
```
TomasTomecek, line 57: How many levels do you suggest for us to use?

TomasTomecek (commented):
Thanks Laura, such a great write-up!

Once we conclude, would be awesome to have a proposal with a solution written here.

celery_tasks_prioritizing/README.md
lbarcziova: I add the suggestion of plan to the doc, I would start with only defining a new queue and see how it will work and maybe later play also with these priority levels, as the documentation regarding it is confusing, found some issues (e.g. https://github.com/celery/celery/issues/4028 - mostly people recommend to emulate the priority on queue level with defining queues for different priorities)

jpopelka (approved):

TomasTomecek (approved):
Thanks for adding the plan, sounds pretty good!

I'm pretty sure this will take some time to polish so splitting the workers per queue sounds like a solid start.

celery_tasks_prioritizing/README.md
```diff
@@ -0,0 +1,105 @@
...
+| run_testing_farm_handler       | run_koji_build_report_handler    |
+| run_propose_downstream_handler | run_testing_farm_results_handler |
+| run_koji_build_handler         | run_installation_handler         |
+| run_pagure_pr_label_handler    | babysit_copr_build               |
+|                                | run_distgit_commit_handler       |
+
+- long-running
+  - tasks that build the SRPM: run_koji_build_handler, run_copr_build_handler
+- short-running:
+  - other tasks
```
jpopelka, line 33: From Martin Pitt:

We do that in Cockpit as well. We have a "webhook" AMQP queue which always gets drained first -- these are high-priority fast tasks such as "react to this pull_request webhook event, interpret it, and put that into the tasks queue", and a "tasks" queue with the slow and bulky stuff ("exercise a test", "build a
VM image").

PR #83, opened 2021-03-29
Dashboard changes

mfocko:
Signed-off-by: Matej Focko <mfocko@redhat.com>

### DIFF
--- a/dashboard-improvements/README.md
+++ b/dashboard-improvements/README.md
@@ -1,0 +1,151 @@
+# Improving dashboard
+
+## Main points
+
+- Informing users about results of SRPM, COPR and Koji builds or Testing Farm
+- Replace service "Build Result" webpage with dashboard
+  - Deprecation of webpage in service
+- Database connection for dashboard
+  - Consider sensitivity of information provided (even with read-only access)
+  - Consider customizing API to the needs of dashboard
+- Consider possible improvements related to either of the points
+
+## Related issues
+
+- [Have a separate page for Testing farm](https://github.com/packit/packit-service/issues/506)
+  - Seems to be at least partially fixed: https://dashboard.stg.packit.dev/results/testing-farm/5603
+  - I guess we need to interlink all related jobs (SRPM, COPR and TF)
+- [IDs for COPR builds in API endpoint (bug)](https://github.com/packit/packit-service/issues/982)
+  - Does not affect usage of COPR build result page so far (is not used), or is the
+    reason for only one chroot result being displayed?
+
+## Current state
+
+- [Already existing TODO list](https://github.com/packit/dashboard/blob/main/TODO.md)
+- We have empty homepage
+  - I can see a potential in showing some statistics, could be helpful to us
+    and also to users.
+  - We could list recently finished jobs or queue.
+- We have implemented result pages
+  - COPR builds:bug mentioned above
+  - Minimal info for testing farm
+- For COPR builds we link directly COPR logs on the dashboard (recent COPR builds),
+  we should probably address our own result page before the COPR one.
+- Navigation:
+  - From each of those we can get back to the branch or the PR
+  - SRPM results -> {can see logs, download SRPM if available}
+  - COPR results -> {SRPM results, COPR logs, compressed build logs from COPR}
+  - TF results -> dead end
+
+## Removal of the _build result_ webpage
+
+0. Fix COPR build bug mentioned above, so that we can see all chroots.
+1. Set dashboard link for all new commit statuses.
+2. Set 2-week long (conveniently matching sprint) grace period for build result
+   page in service.
+
+   _Argument_: 2 weeks is enough for populating commit statuses with links to dashboard
+   and in majority of cases, old links are mainly used for debugging purposes on
+   our side.
+
+3. Remove build result page from service.
+
+   _Alternatively_: Also redirect all requests from service to dashboard in step 1
+   and keep live for longer period of time.
+
+## Providing more information for dashboard through API
+
+### Allowing read-only access to database
+
+- Suspects: `data` in COPR, Koji and Testing Farm
+- Otherwise there does not appear to be anything sensitive present
+- Potential downsides would be:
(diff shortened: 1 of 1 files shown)


dashboard-improvements/README.md
```diff
@@ -0,0 +1,151 @@
...
+- Consider possible improvements related to either of the points
+
+## Related issues
+
+- [Have a separate page for Testing farm](https://github.com/packit/packit-service/issues/506)
+  - Seems to be at least partially fixed: https://dashboard.stg.packit.dev/results/testing-farm/5603
+  - I guess we need to interlink all related jobs (SRPM, COPR and TF)
+- [IDs for COPR builds in API endpoint (bug)](https://github.com/packit/packit-service/issues/982)
+  - Does not affect usage of COPR build result page so far (is not used), or is the
+    reason for only one chroot result being displayed?
```
lachmanfrantisek, line 20: Good, point. I am now not sure if we don't want to unify the behaviour with koji.

dashboard-improvements/README.md
```diff
@@ -0,0 +1,151 @@
...
+   page in service.
+
+   _Argument_: 2 weeks is enough for populating commit statuses with links to dashboard
+   and in majority of cases, old links are mainly used for debugging purposes on
+   our side.
+
+3. Remove build result page from service.
+
+   _Alternatively_: Also redirect all requests from service to dashboard in step 1
+   and keep live for longer period of time.
```
lachmanfrantisek, line 54: How much time do you think this will take?

mfocko: **Implementation?**
Dumping the HTML and adding redirect to it, 1-2 points maybe?

**How long do we want to keep it?**
That is entirely up to us, it is quite interchangeable, e.g.:
`https://stg.packit.dev/srpm-build/3977/logs ~> https://dashboard.stg.packit.dev/results/srpm-builds/3977`

As I said few lines above that, the only problem with switch are old links in statuses... And I'm not sure if anyone checks out the old statuses, unless there are problems present? (In that case it still should not be a problem, since we know about the change and can have a look at it on the dashboard or even API/postgres.

I gave the guess of 2 weeks which seems more than enough for smaller PRs, regarding bigger PRs I would expect the statuses to be replaced.

lachmanfrantisek: Yes, the redirect itself is easy. So if the implementation is no that hard, I would go with this one.

TomasTomecek: agreed with 2/4 weeks for statuses

we could even start with some warning and hint about the work

dashboard-improvements/README.md
```diff
@@ -0,0 +1,151 @@
+# Improving dashboard
+
+## Main points
+
+- Informing users about results of SRPM, COPR and Koji builds or Testing Farm
+- Replace service "Build Result" webpage with dashboard
+  - Deprecation of webpage in service
+- Database connection for dashboard
+  - Consider sensitivity of information provided (even with read-only access)
```
TomasTomecek, line 9: AFAIK only allowlist is sensitive to some extent b/c it contains real logins and namespaces, the rest should be public information

dashboard-improvements/README.md
```diff
@@ -0,0 +1,151 @@
...
+  - Does not affect usage of COPR build result page so far (is not used), or is the
+    reason for only one chroot result being displayed?
+
+## Current state
+
+- [Already existing TODO list](https://github.com/packit/dashboard/blob/main/TODO.md)
+- We have empty homepage
+  - I can see a potential in showing some statistics, could be helpful to us
+    and also to users.
+  - We could list recently finished jobs or queue.
```
TomasTomecek, line 28: yup, sounds like the easiest low-hanging fruit

dashboard-improvements/README.md
```diff
@@ -0,0 +1,151 @@
...
+
+## Related issues
+
+- [Have a separate page for Testing farm](https://github.com/packit/packit-service/issues/506)
+  - Seems to be at least partially fixed: https://dashboard.stg.packit.dev/results/testing-farm/5603
+  - I guess we need to interlink all related jobs (SRPM, COPR and TF)
+- [IDs for COPR builds in API endpoint (bug)](https://github.com/packit/packit-service/issues/982)
+  - Does not affect usage of COPR build result page so far (is not used), or is the
+    reason for only one chroot result being displayed?
+
```
TomasTomecek, line 21: I'd love to add another to the list: allow users to authenticate with their GitHub creds and get a tailred home page with their projects and PRs.

dashboard-improvements/README.md
```diff
@@ -0,0 +1,151 @@
...
+  We could allow retriggering and be sure that user triggering the build again is
+  authorized to do so (spam/DoS protection)
+
+  Cost on our side:
+  Requires changes in service codebase and also we would need to make sure that
+  retriggering cannot be used outside of the dashboard
+
+  - Auth in API `dashboard <=> service`
+  - we could probably restrict IP address of sender to cluster on service side?
```
TomasTomecek, line 89: Require user to be authenticated, disallow retrigger for in-progress jobs - we could use those as roadblocks to avoid disaster.

dashboard-improvements/README.md
```diff
@@ -0,0 +1,151 @@
...
+
+- Personalization of homepage view, e.g. all projects I have _write access_ to or
+  all my open PRs (regardless of write access)
+
+## Additional improvements
+
+- _Badge_ for README with direct link to project's overview page in dashboard ;)
```
TomasTomecek, line 96: :rainbow:

TomasTomecek (commented):
nice, sounds like a solid project :)

dashboard-improvements/README.md
TomasTomecek: okay, this is actually mentioned below

mfocko:
If there are no objections, I would merge. Please react to this comment if it's fine with you :)

lachmanfrantisek (approved):
I am OK with merging this as is -- it's only a result of the SPIKE and we can still discuss if we really want.

PR #87, opened 2021-04-14
Add notes about managing outages

lbarcziova:


outages/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+- IRC - the messages are mirrored to Telegram, this should be enough
+
+### How users are informed
+
+- our [status page](https://dashboard.packit.dev/status) in dashboard
+  - currently we have there only info about API which only checks https://prod.packit.dev/api/healthz
+  - we could provide more info:
+    - manually added info about current outage, have some template: what and when happened, when it will be fixed
+    - history of outages
```
csomh, line 36: Reading through this, I'm wondering whether it would make sense to start using something like [Cachet](https://cachethq.io/).

One more feedback we have received, and maybe would make sense to note here is that the status page should be hosted some other place then the actual project, so that it's possible to communicate the status even if OpenShift is down, for example.

outages/README.md
```diff
@@ -0,0 +1,48 @@
+## How to manage outages
+
+### Who is responsible for resolving outages and what should be done
+
+- rotating role - new / incorporate to an existing one (Sentry master, community master)
```
csomh, line 5: This might not be necessary given a well-developed monitoring and alerting. Until then, "looking for signs" could be the responsibility of the Sentry master.

csomh (commented):
Thanks for this write-up!

One more thing that might be good to consider is to [start using the checks API](https://github.com/packit/packit-service/issues/760#issuecomment-688656092) so that a "neutral" status can be set on PRs. This way we could set a neutral status (instead of an error) for certain known failures of Testing Farm or Copr. We could even make these failures configurable. This would help with generating less noise on PRs when there is a known failure some place.

outages/README.md
jpopelka: +1, would be nice to have some suggested services to use as part of the output of this research. Like the one that Hunor mentioned, which I like.
Bonus: revisit my [One Portal/Platform research](https://github.com/packit/research/tree/main/outage-reporting) I did half a year ago to see whether it's still valid (maybe something's changed with the transition from Portal to Platform?).

lbarcziova: thanks for the suggestions, I like the Cachet as well, I will extend this part and we can then discuss on arch

outages/README.md
```diff
@@ -0,0 +1,55 @@
...
+- identify an outage
+- inform team
+- inform users
+- be responsible for fixing the issue = try to fix it / ask in the team
+- inform when the outage is resolved
```
TomasTomecek, line 10: +1, have a way to provide updates to users, example: https://www.githubstatus.com/incidents/fwtx9db6hz2y

in our case, it could easily be just a github issue

outages/README.md
```diff
@@ -0,0 +1,55 @@
...
+  - mostly info about new exceptions
+  - currently we are informed about each new Sentry issue by mail
+  - [alerts](https://docs.sentry.io/product/alerts-notifications/metric-alerts/)
+    - alert rules can be set up - threshold, notification,
+      [example](https://www.sentry.dev/_assets2/static/fca0872133bf0ed54631d4ba44725879/eb1d2/new-alert-rule-all.png)
+    - alerting is triggered when a threshold is breached
+- Prometheus can provide alerts as well - we would need to collect some metric about failed tasks and create alert rules
+  for that
+- analyze DB info we use in dashboard jobs - periodically check the state of the SRPM builds/ Copr builds/ tests in DB from e.g. the last hour
+  - cron job to check this can be set up
```
TomasTomecek, line 24: I'd like to have all three, thank you very much :)

TomasTomecek: Prometheus will be a great help once we have SLIs implemented :)

outages/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+### How team members are informed
+
+- IRC - the messages are mirrored to Telegram, this should be enough
```
TomasTomecek, line 28: Are you implying one of our current channels? AFAIK OSCI and CKI are routing their alerts to their main channels as well.

outages/README.md
TomasTomecek: +1 Cachet, it also looks like it's simple to deploy: https://github.com/CachetHQ/Docker/blob/main/docker-compose.yml just 2 services

TomasTomecek (approved):
very nice write-up and a ton of work ahead :)

lachmanfrantisek:
> One more thing that might be good to consider is to [start using the checks API](https://github.com/packit/packit-service/issues/760#issuecomment-688656092) so that a "neutral" status can be set on PRs. This way we could set a neutral status (instead of an error) for certain known failures of Testing Farm or Copr. We could even make these failures configurable. This would help with generating less noise on PRs when there is a known failure some place.

We've mentioned this multiple times recently. I'm adding the OGR issue to our Github project so we can start working on it.

lbarcziova:
I am merging this now, we can discuss tomorrow on the architecture and after that, I will create follow-up cards.

PR #89, opened 2021-04-20
Describe caching of git repositories

lachmanfrantisek:
### [Preview of the markdown content](https://github.com/lachmanfrantisek/research/blob/caching-of-git-repos/caching_of_git_repos/README.md)


caching_of_git_repos/README.md
```diff
@@ -0,0 +1,102 @@
...
+## How can this be implemented?
+
+Currently, cloning is done lazily in LocalProject.
+
+1. In case of the very basic version, we can make packit agnostic to this
+   by being able to configure additional arguments for clone commands:
+   _ [packit] Enhance the schema of the user config.
+   _ [packit] Use this value when cloning (LocalProject).
+   _ [deployment] Add persistent volume.
+   _ [deployment] Use `--reference /persistent/volume` in the service config.
```
jpopelka, line 83: Was this supposed to be a bulleted list?
Please check the generated page, not sure this is what you expected.

jpopelka (approved):

caching_of_git_repos/README.md
```diff
@@ -0,0 +1,102 @@
...
+  (We don't need to care about force-pushes, merging, local changes,...)
+  What is fetched counts.
+- We can fetch commits of multiple repositories in the reference repository.
+  - Alternatively, we can use more cache repositories:
+    `git clone --reference-if-able /cache/some-repo-to-clone https://some/repo/to/clone`
+
+## Pros&Cons
+
+- Cloning is much faster for repositories in the cache.
+- Cloning is slower for repositories not present in the cache.
```
TomasTomecek, line 27: we can have a list of repos where we know they are being cached: cockpit, systemd, kernel sound like great candidates, especially when we can reuse the cached copy in upstream and c9s

caching_of_git_repos/README.md
```diff
@@ -0,0 +1,102 @@
...
+- Less memory is needed to clone repositories in the cache.
+  (Which makes it possible to clone kernel for example.)
+- More memory is needed to clone repositories not present in the cache.
```
TomasTomecek, line 30: it depends what would be the workflow to populate the cache but we should do it outside sandcastle/worker

caching_of_git_repos/README.md
```diff
@@ -0,0 +1,102 @@
...
+  - e.g. 1000m needed to clone ogr using cache repo with kernel-ark and systemd (800m has not been enough)
+- Less storage is needed for the cloned repo if it is in the cache.
+  (Only the current state of the repo is saved, historical commits reference the cache repo.)
```
TomasTomecek, line 33: storage is so much cheaper in comparison to memory

caching_of_git_repos/README.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Where to store this cache repository?
+
+- The cache does not need to be writable for cloning.
+  Only for creating/updating.
+- Persistent volumes can be used.
+- How much storage we can afford?
```
TomasTomecek, line 40: our current cost in online is 25€ for 1G of mem and 1€ for 1G of storage - so we can easily start with 16G of storage

caching_of_git_repos/README.md
```diff
@@ -0,0 +1,102 @@
...
+  - Storage is cheap, but git repositories can be really big after some time.
+  - I haven't tested efficiency of the scenario with a lot of repositories in one cache repository.
+- Data itself can be shared between stage/prod if we want. (Probably not wanted and hard to do in openshift.)
+
+## How to create the cache repository?
+
+- Manually on request. Mount the volume once with more memory and fetch the needed repository.
+- Manually on sentry issue. As previous but gather the problematic repos in sentry.
+- Start with kernel manually and add new ones on the go.
```
TomasTomecek, line 49: It would be delightful if we could create a workflow how to populate the cache (e.g. regen it weekly), have metadata which repositories are in the cache and when should they be "attached" so they are being used transparently.

caching_of_git_repos/README.md
```diff
@@ -0,0 +1,102 @@
...
+
+## What repositories we want there?
+
+- Just kernel.
+- A group of hardcoded/configured repositories.
+- All repositories matching some condition (at least some commits, some size, ...)
+- All repositories. (Add if not present.)
```
TomasTomecek, line 56: using kernel for a PoC would be very nice, we could prototype this in the SIG

after it's proven, we could introduce this in the upstream to projects with at least N runs a week?

caching_of_git_repos/README.md
```diff
@@ -0,0 +1,102 @@
...
+   _ [packit] Enhance the schema of the user config.
+   _ [packit] Use this value when cloning (LocalProject).
+   _ [deployment] Add persistent volume.
+   _ [deployment] Use `--reference /persistent/volume` in the service config.
+2. Alternatively, we can have the cache directory configurable (optionally).
+3. If we want more clever behaviour, we need to put the cloning logic to packit.
+4. Or, we can forward some method for handling the cloning.
+   (Defined in the service repo, run in the packit.)
+
+## Is this relevant for the CLI users?
```
TomasTomecek, line 89: I'm sorry but I don't see any value here since people already have those projects cloned.

TomasTomecek (commented):
well written!

caching_of_git_repos/README.md
lachmanfrantisek: Prettier did not like my old version..;( Thanks

caching_of_git_repos/README.md
lachmanfrantisek: I agree that it does not make sense to spend much time on it, but (as I wrote below, in this part),

* second repo (upstream/downstream) is temporarily cloned by default
* both repos are temporarily cloned when URL is used as an argument (That's what I use for example. You don't need to care about the state of the local repository.)

TomasTomecek: I really liked how you explained it during the arch meeting, it indeed makes perfect sense for those temporary clones, +1

PR #91, opened 2021-04-28
Ideas about splitting the source-git work and upstream work

lachmanfrantisek:
### [Preview of the markdown content](https://github.com/lachmanfrantisek/research/blob/split-the-stream/split-the-stream/README.md)

### DIFF
--- a/split-the-stream/README.md
+++ b/split-the-stream/README.md
@@ -255,3 +255,35 @@ Some related GitLab issues:
    - We can define pipelines or use commit statuses (=detached pipelines).
 4. How are CI results going to be displayed in dist-git MRs? We wan't to know this so that we can think about ways to take those results and display them for contributors on the source-git MRs.
    - Displayed as pipelines.
+
+## The plan
+
+1. Set up a new repository for the stream worker.
+   - `stream-worker`/`packit-stream-worker`/`source-git-worker`/...
+   - Build the image in quay.
+   - Setup zuul and pre-commit.ci.
+   - Create a `stable` branch.
+2. Set up a new deployment repository for stream.
+   - Create the new playbooks and share as much as possible with current workflow.
+   - We will have only one deployment for stream for start.
+   - Increase (=buy) the resources in openshift online.
+   - Deploy the stream service to openshift online.
+   - Update the script for moving `stable` branches.
+3. Implement the stream worker.
+   - New celery tasks are defined.
+   - Implementation is done as new handlers.
+   - Start with the really basic version so we can work on deployment ASAP:
+     - _If user creates a merge-request on the source-git repository, create a matching merge-request to the dist-git repository._
+4. Put the current worker away from the service.
+   - `packit-worker`/`packit-upstream-worker`/...
+   - What about the `process_message` task? (SPIKE card for that has been created.)
+     - Do we want to share it? What about a dedicated worker just for this?
+   - Move the build process.
+   - Setup zuul and pre-commit.ci.
+   - Update deployment if necessary.
+   - Update the script for moving `stable` branches.
+   - (Can be done in parallel with other steps.)
+5. Implement _Sync the CI results from the dist-git merge-request to the source-git merge-request._
+6. Implement _If the dist-git is updated, update the source-git repository by opening a PR._
+7. Implement _If the source-git merge-request is updated, update the dist-git merge-request._
+8. Implement _If the source-git merge-request is closed, close the dist-git merge-request._
--- a/split-the-stream/README.md
+++ b/split-the-stream/README.md
@@ -91,6 +91,48 @@ We have multiple options:
 - To share the code between the upstream and source project, we need to create some shared libraries.
   - Can lead to 3. and/or having another project on our dependency chain.
 
+### Database
+
+What are the differences in the schema?
+
+- Event related models can be shared (`GitProject`, `JobTriggerModel`, `RunModel`, PullRequestModel`, ...).
+- Result models are probably not necessary (`SRPMBuildModel`, `CoprBuildModel`, ...).
+- Allow/deny list can be preserved.
+- Models for the setup procedure aren't necessary for stream (`InstallationModel`, `ProjectAuthenticationIssueModel`).
+- Connection between source-git and dist-git MRs can be done by creating a new join table.
+- If we need to track the results, we need to create a new model for that.
+
+We have multiple ways how we can manage database schema after the split:
+
+#### One schema for all
+
+- One schema that works for all use-cases.
+- The schema is defined in one place.
+- Databases can contain unused tables.
+- Schema can be more complicated.
+
+#### Independent schemas
+
+- Each schema fits the use-case.
+- Common changes are harder to share.
+
+#### Multiple alembic branches
+
+- https://alembic.sqlalchemy.org/en/latest/branches.html
+- Each use-case has its own alembic branch.
+- The definition is in one place.
+- Does not help with sharing of changes. (There is only merge and no rebase available.)
+
+#### Multiple alembic bases
+
+- https://alembic.sqlalchemy.org/en/latest/branches.html#working-with-multiple-bases
+- Multiple independent alembic migration paths. (We work on multiple branches in parallel.)
+- Different base can be located in a different location.
+  - There can be one shared schema and each use-case (upstream/CentOS-stream/?Fedora) can have a second one.
+- Ideally, bases are independent, but it is possible to
+  [specify dependencies](https://alembic.sqlalchemy.org/en/latest/branches.html#branch-dependencies)
+  between revisions.
+
 ## Linking of the merge-requests
 
 We have two goals:
--- a/split-the-stream/README.md
+++ b/split-the-stream/README.md
@@ -105,6 +105,11 @@ We have two goals:
   (Mapping would be problematic.)
 - For the purpose of the service, we can save the pairs to the database.
 
+Some related GitLab issues:
+
+- [Rearchitect MR widget mergeability logic](https://gitlab.com/gitlab-org/gitlab/-/issues/300042)
+- [Rearchitect / Refactor MR Widget [Discovery]](https://gitlab.com/gitlab-org/gitlab/-/issues/324381)
+
 ## Merge trains
 
 Allow merging multiple MRs in one target branch safely:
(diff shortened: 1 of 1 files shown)


split-the-stream/README.md
```diff
@@ -0,0 +1,89 @@
...
+  - Sync the CI results from the dist-git merge-request to the source-git merge-request.
+- If the dist-git is updated, update the source-git repository.
+
+Should have:
+
+- If the source-git merge-request is updated, update the dist-git merge-request.
+
+Could have:
+
+- User is able to re-trigger the dist-git CI from the source-git merge-request.
```
lachmanfrantisek, line 20: @csomh @TomasTomecek Can you please verify this part so I can continue with the spike?

split-the-stream/README.md
```diff
@@ -0,0 +1,89 @@
...
+Some ideas, possibilities, pros and cons of moving out the source-git related work.
+
+## What does the source-git workflow mean?
+
+Must have:
+
+- If user creates a merge-request on the source-git repository:
+  - Create a matching merge-request to the dist-git repository.
+  - Sync the CI results from the dist-git merge-request to the source-git merge-request.
+- If the dist-git is updated, update the source-git repository.
```
csomh, line 12: I would add: "... by opening a PR", or something similar. The main idea would be that the service should never directly push to a branch. All interactions with the repositories (both source-git and dist-git) should happen through PRs.

split-the-stream/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+Should have:
+
+- If the source-git merge-request is updated, update the dist-git merge-request.
```
csomh, line 16: Additionally: if the source-git MR is closed, close the dist-git MR—to make the mirroring complete.

csomh: I'm not sure, whether there should be a mechanism when syncing the MRs to check that the dist-git MR was not modified since the last update, so that dist-git MRs edited by contributors are not overwritten.

Furthermore:
There should be means to trigger or re-do the update of dist-git MRs both through the service (with commands in comments) and through using the CLI.

split-the-stream/README.md
csomh: Done.

lachmanfrantisek: Thanks.

split-the-stream/README.md
lachmanfrantisek: Also for non-user dist-git updates? (Contributor's merge-request is merged to dist-git or mass rebuild or other external change.

split-the-stream/README.md
lachmanfrantisek: > through using the CLI.

Will this command change the merge request or the local dist-git repository?

split-the-stream/README.md
csomh: That's a good point. I would say, yes, even if in the end, those source-git MRs are merged automatically.

split-the-stream/README.md
csomh: For the sake of simplicity it should update a local dist-git repo. It would fall on the developer to push those changes to update the MR.

split-the-stream/README.md
lachmanfrantisek: Updated.

split-the-stream/README.md
lachmanfrantisek: Updated.

split-the-stream/README.md
```diff
@@ -0,0 +1,178 @@
...
+- We need to maintain two separate projects. (We need more people or reduce the productivity.)
+- We are not motivated to improve the current projects.
+- To share the code between the upstream and source project, we need to create some shared libraries.
+  - Can lead to 3. and/or having another project on our dependency chain.
+
+## Linking of the merge-requests
+
+- We can set the dependent merge-request across multiple projects
+  (e.g. https://gitlab.com/lachmanfrantisek/kernel-ark/-/merge_requests/1)
+  - I couldn't find a related API.
```
mfocko, line 98: You couldn't have, it's just in the UI:
https://gitlab.com/gitlab-org/gitlab/-/issues/12551

split-the-stream/README.md
```diff
@@ -0,0 +1,178 @@
...
+    (The details does not show anything relevant:
+    [example response](https://gitlab.com/api/v4/projects/23847254/merge_requests/1).)
+- Mentioning the other merge-request is always a possibility.
```
mfocko, line 101: What about protecting the branch?
(But that also moves the full responsibility for branches to us... just maintainers and packit maybe?)

mfocko:
So I understood correctly from documentation and this PR that GitLab's pipeline is tied to their own CI with no way of bending it?

split-the-stream/README.md
lachmanfrantisek: Thanks for the link!

split-the-stream/README.md
lachmanfrantisek: Aha, you mean to not allow the merge?

I think there are two goals with this (I'll update the text to cover that.):
* User can easily find the related merge request on the web UI. (In both ways.)
* We can get the related merge request from the service.

lachmanfrantisek:
> So I understood correctly from documentation and this PR that GitLab's pipeline is tied to their own CI with no way of bending it?

I would say so. I'll try to find some existing issues for that.

lachmanfrantisek:
OK, based on [this](https://gitlab.com/gitlab-org/gitlab/-/issues/20907#note_300399873) there is a way to do that (if I get that right). I'll play with that more.

split-the-stream/README.md
```diff
@@ -0,0 +1,184 @@
...
+Some ideas, possibilities, pros and cons of moving out the source-git related work.
+
+## What does the source-git workflow mean?
+
+Must have:
+
+- If user creates a merge-request on the source-git repository:
+  - Create a matching merge-request to the dist-git repository.
+  - Sync the CI results from the dist-git merge-request to the source-git merge-request.
+- If the dist-git is updated, update the source-git repository by opening a PR.
```
TomasTomecek, line 12: I'd say this only applies to the specfile since synchronizing patches to the source-git repo would be immensely hard.

split-the-stream/README.md
```diff
@@ -0,0 +1,184 @@
...
+1. Should we start developing a new service or modify the existing packit-service to be able to deploy only with a gitlab-endpoint?
+2. How to link merge requests in the src namespace to the ones in the rpms namespace using the GitLab API? This should be bidirectional.
+3. Check the GitLab API to learn more about working with merge-trains and pipelines, in order to support the UX for merging source-git MRs (see the doc linked bellow).
+4. How are CI results going to be displayed in dist-git MRs? We wan't to know this so that we can think about ways to take those results and display them for contributors on the source-git MRs.
+
+## Split
+
+We have multiple options:
+
+### 0. no split
```
TomasTomecek, line 36: the biggest downside there is that an outage in the upstream or fedora workflow would cause outage in the Stream workflow

split-the-stream/README.md
```diff
@@ -0,0 +1,184 @@
...
+### 1. same codebase, new deployment
+
+- No extra cost of maintenance of two codebases.
+- New jobs are implemented as new handlers.
+- Different identities can be used in one git forge (=gitlab.com).
+- Resources can be tweaked separately.
+- Fedora-source-git friendly: easy combination of events (different for Fedora and Stream)
+  and handlers (=implementation, can be shared).
+
+### 2. separate workers
```
TomasTomecek, line 54: this look really intriguing to be honest - those workers could then have a lifecycle on their own

split-the-stream/README.md
```diff
@@ -0,0 +1,184 @@
...
+- We can get rid of the old bad staff in the packit-service.
+- We can go through the same pain we've already gone through.
+- We need to maintain two separate projects. (We need more people or reduce the productivity.)
+- We are not motivated to improve the current projects.
+- To share the code between the upstream and source project, we need to create some shared libraries.
+  - Can lead to 3. and/or having another project on our dependency chain.
+
+## Linking of the merge-requests
+
+We have to goals:
```
TomasTomecek, line 96:
```suggestion
We have two goals:
```

TomasTomecek (commented):
Wow, such detailed research! thank you; we have lots to discuss here

split-the-stream/README.md
```diff
@@ -0,0 +1,184 @@
...
+
+Should have:
+
+- If the source-git merge-request is updated, update the dist-git merge-request.
+- If the source-git merge-request is closed, close the dist-git merge-request.
+
+Could have:
+
+- User is able to re-trigger the dist-git CI from the source-git merge-request.
+- User is able to re-create the dist-git CI from the source-git merge-request.
```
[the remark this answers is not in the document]
lachmanfrantisek: I wanted to write `re-create the dist-git MR`. (Just create a fresh MR from the source-git one. I recall we've discussed this functionality.)

split-the-stream/README.md
```diff
@@ -91,6 +91,39 @@ We have multiple options:
...
+- The schema is defined in one place.
+- Databases can contain unused tables.
+- Schema can be more complicated.
+
+#### Independent schemas
+
+- Each schema fits the use-case.
+- Common changes are harder to share.
+
+#### Multiple alembic branches
```
TomasTomecek, line 110: it seems this would be a nightmare to maintain - if we only have 2 repos, 'one schema for all' seems like the easiest to maintain

split-the-stream/README.md
```diff
@@ -0,0 +1,243 @@
...
+- The new service can be more lightweight and efficient.
+- We can iterate on the prototype more quickly.
+- We can get rid of the old bad staff in the packit-service.
+- We can go through the same pain we've already gone through.
+- We need to maintain two separate projects. (We need more people or reduce the productivity.)
+- We are not motivated to improve the current projects.
+- To share the code between the upstream and source project, we need to create some shared libraries.
+  - Can lead to 3. and/or having another project on our dependency chain.
+
+### Database
```
csomh, line 94: I'm wondering what would be the common parts to be shared between the Packit and source-git use-cases? And how much would they represent from the total schema definition for the Packit and source-git use-case?

In the source-git usecase we would need to store data for the following types of pipelines:
- pipelines triggered by source-git MRs, which will create/update dist-git MRs;
- pipelines triggered by dist-git MRs being merged, which will merge the corresponding source-git MRs;
- pipelines triggered by other dist-git changes, which will open source-git MRs.

Of course the functionality above might change, based on what solutions we choose, but nevertheless the above could be used to tell what parts of the current schema could be used to support the above use cases. At first sight, it seems to me, there is no commonality, but I might be terribly wrong about this :smiley:

lachmanfrantisek: Thanks for the comment. I've covered this in the text and put it on the agenda also for the next arch. meeting so we can agree on something.

lachmanfrantisek:
We went through that on Tuesday and improved&agreed on the plan. Let's merge it.

PR #100, opened 2021-06-15
Add notes about deployment and testing strategies

lbarcziova:


jpopelka (approved):
Thanks!

Can we call "our" stg/prod deployment a Blue-Green or how is it different?
Canary and A/B look very similar to me, the main difference I see is
- a percentage of requests is sent to "Canary"
- a subset (based on a condition) of requests is sent to "experiment" (A/B)
and different focus (performance & error rate vs. experimentation)?

csomh:
I know this is still WIP and you might have already planned doing this, but nevertheless... One thing I'm curious to learn from this research is to learn what are our technical options to implement these strategies (for example: does OpenShift help with this?) and what would it take (how much work/effort/risk) to get there.

lbarcziova:
> I know this is still WIP and you might have already planned doing this, but nevertheless... One thing I'm curious to learn from this research is to learn what are our technical options to implement these strategies (for example: does OpenShift help with this?) and what would it take (how much work/effort/risk) to get there.

Yes, I wanted to include some of these things, good point.

TomasTomecek:
> I know this is still WIP and you might have already planned doing this, but nevertheless... One thing I'm curious to learn from this research is to learn what are our technical options to implement these strategies (for example: does OpenShift help with this?) and what would it take (how much work/effort/risk) to get there.

OpenShift 4 ideally so we'd have another reason drop online and get a 4 cluster finally.

deployment-verification/README.md
```diff
@@ -0,0 +1,93 @@
...
+- different versions of the same service run simultaneously as “experiments” in the same environment for a period of time
+- released to subset of users under specific condition
+- primarily focused on experimentation and exploration
+- for testing new features
+
+---
+
+- probably not really relevant to us
+
+### Shadow deployment
```
TomasTomecek, line 71: wow, this would be mind-blowing although I agree it would take too much work to do it right

deployment-verification/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+| Deployment or testing pattern                                                                                     | Zero downtime | Real production traffic testing | Releasing to users based on conditions | Rollback duration |
+| ----------------------------------------------------------------------------------------------------------------- | ------------- | ------------------------------- | -------------------------------------- | ----------------- |
+| Blue/green (Version 2 is released alongside Version 1; the traffic is switched to Version 2 after it is tested. ) | ✓             | x                               | x                                      | Instant           |
+| Canary (Version 2 is released to a subset of users, followed by a full rollout.)                                  | ✓             | ✓                               | x                                      | Fast              |
+| A/B (Version 2 is released, under specific conditions, to a subset of users.)                                     | ✓             | ✓                               | ✓                                      | Fast              |
```
TomasTomecek, line 92: A/B looks like the clear winner :)

deployment-verification/README.md
```diff
@@ -0,0 +1,93 @@
+# Deployment and testing strategies
+
+### Blue-Green deployment
```
csomh, line 3: My understanding of the blue-green deployment is that we would start with 2 environments both having v1 of the service deployed. Then we would release v2 to the blue environment, test and verify it, and if it seem alright, redirect all service trafic from the green env to the blue one. This is when the colors would "switch", so that the env serving live trafic is always called green, and the one where upgrades and verification should be done is called blue.

Could you maybe update the description to better reflect the above?

TomasTomecek (approved):
I wonder what the next steps will be here.

lbarcziova:
> I wonder what the next steps will be here.

We can at least go through these strategies on the architecture and think whether we want to use some process from these in the future, WDYT?

PR #112, opened 2021-09-06
GitHub Actions

mfocko:
Signed-off-by: Matej Focko <mfocko@redhat.com>

### DIFF
--- a/github-actions/README.md
+++ b/github-actions/README.md
@@ -1,0 +1,33 @@
+# GitHub Actions
+
+## Benefits and options for Packit and users of Packit
+
+_TODO_
+
+## Disadvantages and caveats
+
+_TODO_
+
+## How can we use it
+
+_TODO_
+
+### Rough estimation of time required
+
+_TODO_
+
+## Sharing the work done with GitLab (and potentially Pagure)
+
+_TODO_
+
+## Testing Farm
+
+_TODO_
+
+### Running tests in GitHub Action
+
+_TODO_
+
+### Testing farm via API from GitHub Action
+
+_TODO_


github-actions/README.md
```diff
@@ -0,0 +1,87 @@
...
+- No cache is present, which also introduces:
+  - slower build time
+  - potential time out, since Actions are time limited (iirc 24 hrs per action)
+
+### Security concerns
+
+In case of utilizing API via GitHub Actions there is a risk of modifying actions that could result
+in leaking of the secrets. GitHub mitigates this by censoring the output of GitHub Actions, but with
+our use-case it would be still possible to (indirectly) modify scripts that are run and leak the
+secrets.
```
praiskup, line 35: Secrets shouldn't be available during pull-request workflows ... and push workflows are fully under the maintainer's control.

github-actions/README.md
```diff
@@ -0,0 +1,87 @@
...
+- Allows user to **easily** customize the dependencies, i.e. does not require changes to be made to
+  the Sandcastle image.
+- Provides access to created SRPMs without dependency on Copr or Packit Service.
+
+## Disadvantages and caveats
+
+### SRPM builds
+
+- If we were to trigger Copr builds with SRPM created in GitHub Action we introduce blocking action of
+  acquiring the SRPM from GitHub.
```
TomasTomecek, line 25: we could potentially trigger p-s' API from the action and POST the SRPM

would be interesting to discuss alternative solutions for this step

github-actions/README.md
```diff
@@ -0,0 +1,87 @@
...
+- No cache is present, which also introduces:
+  - slower build time
+  - potential time out, since Actions are time limited (iirc 24 hrs per action)
```
TomasTomecek, line 28: this is a good point, since job scheduling would be completely out of our control

github-actions/README.md
```diff
@@ -0,0 +1,87 @@
...
+- do not run the action if user has no write-access (enforces check by the maintainer)
+  - differentiating «maintainer approved»?
+  - how to decide if changes can be evaluated or not? (i.e. user is trusted to modify the files)
+    - list of files that cannot be modified externally? However it might be required to get those
+      from the default branch, since branch where changes are merged may be compromised.
+
+### Versioning
+
+In case we do not provide specialized images on which Packit can be run, we can run into issues with
+different versions of Packit being used (RPM vs PyPI vs images).
```
TomasTomecek, line 49: I'm afraid we'd need to provide packit in a container where the env is in full control from us

github-actions/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+_TODO_
+
+#### Running tests in GitHub Action
+
+_TODO_
+
+#### Testing farm via API from GitHub Action
+
+_TODO_
```
lachmanfrantisek, line 79: Ideally, leave this to the Testing Farm team (=Miro) and allow some integration.

github-actions/README.md
lachmanfrantisek: This is handy but will probably cause problems with the Copr builds for PRs. How we can do this?

mfocko: Ad pull request workflow, they disable secrets by default (with an override): https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#potential-impact-of-a-compromised-runner

In ideal case I would like to split the action to multiple parts and pass the secrets to the environment only for calling packit that directly queries API.

Follow-up question:
- [ ] Do we **always** create SRPM for `packit copr-build`, **even if** there is an SRPM already present?

lachmanfrantisek: >     * Do we **always** create SRPM for `packit copr-build`, **even if** there is an SRPM already present?

I think so, but if we want to change this and reuse the existing srpm, this is easy to do. Maybe as another CLI argument?

LecrisUT:
I want to point out an important use-case abouy this. This is necessary when upstream uses release actions, e.g. uploading to pypi. In this case it creates a race condition with `propose_downstream`. Using a Github action approach we can have better control of when to trigger these builds

PR #113, opened 2021-09-13
Add requirements for the spec-file library

lachmanfrantisek:
Signed-off-by: Frantisek Lachman <flachman@redhat.com>

### DIFF
--- a/specfile/README.md
+++ b/specfile/README.md
@@ -1,0 +1,81 @@
+# Packit requirements for the spec-file library
+
+## Manipulation
+
+- We create a Python object by specifying a path of the spec-file and (optionally) path to spec-file sources
+  (the directory with the spec-file).
+- (optional) After a change, we can reload the content (can be done by recreating of the object).
+- After the manipulation, the content is synced to the file or, we have a way to do this explicitly.
+- Be able to copy a section of spec-file to another object.
+- When reading the values, we want to be able to have both version with macros expanded
+  and raw without expansion.
+- When changing the spec-file, we want to operate on the raw spec-file.
+  - All changes results in a minimal and local diff.
+
+## Version and release
+
+- The version and release fields can be get.
+- We can set the new version. If release is not specified, reset the release number to 1.
+
+## Changelog
+
+- We can get the content of the changelog.
+- We can add a new changelog entry:
+  - By providing only the changelog text.
+  - Optionally, field like `name` or `email` can be specified.
+    (Defaults of bumbspec are used by default.)
+
+## Sources
+
+- We can get the sources and patches.
+- We can download the sources from a lookaside cache.
+- We can download remote sources (sources in form of URL).
+- We can change the value of a source.
+- We can get archive name for source.
+
+## Comments
+
+- For patches and other fields, we want to add a comment without a need to manipulate with the file directly.
+
+  - What about having a get/set `comment` value for various spec-file attributes for comment lines right above the item without any non-comment line?
+
+    ```
+    # this is
+    # the comment
+    # for the first source
+    Source1: source1.tar
+
+    # this is not a comment for the second source
+
+    Source2: source2.tar
+    ```
+
+## Patches
+
+- We can get the patches.
+  - Including the comment above (see the section above).
+- We can add a new patch:
+  - Automatically pick the right number.
+  - Check if the patch isn't already present.
+  - If there already are patches, then the patch is added after them.
+  - If there are no existing patches, the patch is added after Source definitions
+  - Be able to set a comment used above the patch line (see the section above).
+- Concept of applied patches. (Probably source-git specific.)
+  - Behave like a commented patch.
+  - We are able to list the applied patches.
+  - We can apply the patches = comment the patch.
+- We can remove all the patches.
+
+# Other content
+
+- We can get and replace the method used in `prep` section (e.g. `autosetup`, `setup`, `autopatch`,...).
+- Get URL tag.
+
+# Workflows (can be implemented elsewhere but would be nice)
+
+- Bumbspec: change version and add a new changelog entry.
+- Be able to copy content of the whole spec-file to another object and be able to ignore a section during that (e.g. whole content but the changelog).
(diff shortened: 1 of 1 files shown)


specfile/README.md
```diff
@@ -0,0 +1,75 @@
+# Packit requirements for the spec-file library
+
+## Manipulation
+
+- We create a Python object by specifying a path of the spec-file.
```
TomasTomecek, line 5: I know that we have to specify the path to spec file sources - should we mention that?

specfile/README.md
```diff
@@ -0,0 +1,75 @@
...
+- (optional) After a change, we can reload the content (can be done by recreating of the object).
+- After the manipulation, the content is synced to the file or, we have a way to do this explicitely.
+- Be able to copy a section of spec-file to another object.
+
+## Version and release
+
+- The version and release fields can be get.
+- We can set the new version. If release is not specified, reset the release number to 1.
```
TomasTomecek, line 13:
```suggestion
- We can set the new version. If release is not specified, reset the release number to 1 while preserving all the macros.
```

TomasTomecek (approved):

specfile/README.md
lachmanfrantisek: What about mentioning this in the beginning that we want to save only relevant changes, not the expanded macros.

specfile/README.md
```diff
@@ -0,0 +1,75 @@
+# Packit requirements for the spec-file library
+
+## Manipulation
+
+- We create a Python object by specifying a path of the spec-file.
+- (optional) After a change, we can reload the content (can be done by recreating of the object).
+- After the manipulation, the content is synced to the file or, we have a way to do this explicitely.
```
csomh, line 7:
```suggestion
- After the manipulation, the content is synced to the file or, we have a way to do this explicitly.
```

specfile/README.md
```diff
@@ -0,0 +1,75 @@
...
+- Be able to copy a section of spec-file to another object.
+
+## Version and release
+
+- The version and release fields can be get.
```
csomh, line 12: Do you mean this with macros expanded or the raw values? Or both?

specfile/README.md
```diff
@@ -0,0 +1,75 @@
...
+- We can set the new version. If release is not specified, reset the release number to 1.
+
+## Changelog
+
+- We can get the content of the changelog.
+- We can add a new changelog entry:
+  - By providing only the changelog text.
+  - Optionally, field like `name` or `email` can be specified.
```
csomh, line 20: What values are going to be used, if no name and email is specified?

specfile/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+- We can get the sources and patches.
+- We can download the sources from a lookaside cache.
+- We can download remote sources (sources in form of URL).
+- We can change the value of a source.
+- We can get archive name = basename of the first Source in SPEC file a.k.a. `Source`/`Source0`.
+
+## Comments
+
+- For patches and other fields, we want to add a comment without a need to manipulate with the file directly.
```
csomh, line 32: :+1:

specfile/README.md
```diff
@@ -0,0 +1,75 @@
...
+  - By providing only the changelog text.
+  - Optionally, field like `name` or `email` can be specified.
+
+## Sources
+
+- We can get the sources and patches.
+- We can download the sources from a lookaside cache.
+- We can download remote sources (sources in form of URL).
+- We can change the value of a source.
+- We can get archive name = basename of the first Source in SPEC file a.k.a. `Source`/`Source0`.
```
csomh, line 28: What about the others? Why just the first one?

csomh (commented):
I think it should also be clarified for each use-case (or in general) if we want those operations with macros expanded or on the raw text in the spec-file.

specfile/README.md
lachmanfrantisek: I've got that from the docstring of the rebase-helper method..;) I'll remove the `first`...

specfile/README.md
lachmanfrantisek: This is with what I am really not sure in general -- do we want to have both modes?

For us, it's crucial to be able to do the edits without side effects (the diff after a change is local and minimal).

specfile/README.md
lachmanfrantisek: I've added `All changes results in a minimal and local diff.` to the very beginning.

specfile/README.md
lachmanfrantisek: Is this right (mainly the default)?
```
- We create a Python object by specifying a path of the spec-file and (optionally) path to spec-file sources
  (the directory with the spec-file).
```

lachmanfrantisek:
@TomasTomecek @csomh I've included your suggestion but am still not sure about the macros.

What do you think about that?

Here are my points:
* Not expanding macros can lead to non-parsable spec-file.
* Expanding macros can make it hard to edit the file with a minimal diff.
* I am thinking about supporting both modes but making it read-only when macros are expanded. Or just clearly state that the final diff can be weird.

csomh:
> @TomasTomecek @csomh I've included your suggestion but am still not sure about the macros.

Can we think about this topic as: *we want an API to be able to edit spec-files similar to how package developer do it*?

If I put it this way, than in some cases—mostly when editing the files (?)—the API should provide access to the raw spec-file, as this is what a package developer would also do. 

In other cases—mostly when reading the values of tags (?)—we are interested in the values with macros expanded (reading `Source0` for example which often contains macros).

Following the train of though above I would say that the methods which edit the spec-file should operate on the raw-spec, eventually doing some kind of validation after editing that the spec still can be parsed. Reading methods should have an option to either get things with macros expanded or not. The latter would be probably handy when implementing the methods to edit the spec.

lachmanfrantisek:
@csomh I agree, both modes makes sense only for reading. I've included this in the document. I hope we are done for now and can go deeper together with other teams.

csomh (approved):

PR #116, opened 2021-09-23
Add proposal to support monorepos in source-git

csomh:

### DIFF
--- a/source-git-monorepos/README.md
+++ b/source-git-monorepos/README.md
@@ -1,0 +1,70 @@
+# Support monorepos in source-git
+
+Monorepos store the source code of multiple projects in a single
+repository. They are a special use-case from a source-git point of view,
+because an upstream monorepo maps to multiple dist-git repositories in the
+distribution.
+
+One example for the above is [LLVM] from which code flows into multiple
+dist-git repositories in Fedora Linux.
+
+Currently the source-git commands in source-git map an upstream repo to a
+single dist-git repo in the distribution.
+
+To change this, a mechanism to specify which path(s) in the monorepo to map to
+which dist-git repo in the distribution should be introduced.
+
+This could be done by introducing a list-type `packages` key. Items of the
+list are objects with keys identical to the current top-level keys in
+`source-git.yaml`, except `upstream_project_url` and `upstream_ref`—these
+are left to be top-level keys only, and a new `package_paths` key which is a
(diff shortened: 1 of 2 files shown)


jpopelka (approved):

TomasTomecek (approved):

kwk:
LGTM

kwk (approved):
LGTM

source-git-monorepos/README.md
```diff
@@ -0,0 +1,70 @@
...
+`downstream_package_name`. Packit limits patch generation to these paths.
+
+Example:
+
+```yaml
+upstream_project_url: https://github.com/llvm/llvm-project.git
+upstream_ref: llvmorg-13.0.0-rc3
+packages:
+  - downstream_package_name: python-lit
+    specfile_path: .distro/python-lit/python-lit.spec
```
tstellar, line 31: We get the sources for python-lit from pypi, so I don't know if we need to get our sources from the monorepo for this one.

kwk: @tstellar but I do build python-lit all from the `<llvm-root-dir>/utils/lit` tree and I can recall that we ran into a situation when python-lit was too outdated on pypi to build or test the latest tests in the upstream llvm tree when doing a snapshot build of LLVM.

serge-sans-paille: From the fedora point of view, I think it's better to build from PyPI, because that's the official place where the package is released.

kwk: @serge-sans-paille I agree for official releases but not for snapshots for the reasons I've explained. But I'm sure we can work this out without jeopardising the one or the other.

csomh: Thank you all! I made a note to consider this (the ability to choose where to get source archives from) going forward.

stefwalter:
> > One of the fundamental bits of "magic" about source-git is the ability to take a downstream commit and do a PR/MR back upstream in the obvious standard ways. Do monorepos allow that? I wasn't able to determine from the research.
> 
> Yes, they allow that. The source-git design doesn't change here, in the sense that an upstream repo has a corresponding source-git repo, which has downstream changes as Git commits. The only difference with monorepos is that a single source-git repo maps to multiple dist-git repos. The research proposed a way to configure this mapping in source-git.

That makes so much sense. There's so many cases where multiple packages come from a single upstream project. Sorry for misunderstanding.

PR #120, opened 2021-10-11
research making decisions

TomasTomecek:
TODO
- [ ] Pick a direction we want to go with.
- [ ] Define guidelines.
- [ ] Try to make a decision on next arch using the guidelines.


lachmanfrantisek (commented):
This article about `Teal organisations` looks relevant: https://reinventingorganizationswiki.com/theory/decision-making/ especially the part about `Consent-based decision making`.

making-decisions/README.md
```diff
@@ -0,0 +1,85 @@
...
+proposal ('duh). They advise for the decision maker to make a call based on
+research, advices and being the subject matter expert. I instead suggested to
+have a consensus on a decision. The argument from the article is that reaching
+consensus is too expensive time-wise and resource-wise. I'd still try what I
+propose and reevaluate what's the most optimal process for us after a period of
+time.
+
+We can definitely still learn from the article so I strongly suggest reading it.
+
+## [Consent Decision Making](https://thedecider.app/consent-decision-making)
```
csomh, line 71: I like this the most, and something similar is also mentioned in the Teal organization description above.

It gives room for discussions (which the current proposal doesn't mention how they should happen) and makes objections a part of the process, while also making the whole thing a little bit faster by not requiring full consensus.

PR #126, opened 2021-11-04
Add plan how to start building SRPMs in Copr

lbarcziova:


building-srpm-in-copr/README.md
```diff
@@ -90,3 +85,27 @@ copr_client.build_proxy.create_from_custom(ownername="lbarczio", projectname="og
                            script_builddeps=["git", "packit", "python3-wheel", "python3-pip", "python3-setuptools",
                             "python3-setuptools_scm", "python3-setuptools_scm_git_archive"])
 ```
+
+### Plan how to build SRPMs in Copr in Packit Service
+
+- create CLI command or only Packit API method for preparing sources (would mostly use existing functionality, described above)
```
TomasTomecek, line 91: such command would be also really helpful when testing out things and especially when configuring the archive generation

building-srpm-in-copr/README.md
```diff
@@ -90,3 +85,27 @@ copr_client.build_proxy.create_from_custom(ownername="lbarczio", projectname="og
                            script_builddeps=["git", "packit", "python3-wheel", "python3-pip", "python3-setuptools",
                             "python3-setuptools_scm", "python3-setuptools_scm_git_archive"])
```
TomasTomecek, line 86: I'm wondering if Copr can actually read the BRs in the spec and install them without us passing this list

building-srpm-in-copr/README.md
```diff
@@ -90,3 +85,27 @@ copr_client.build_proxy.create_from_custom(ownername="lbarczio", projectname="og
...
 ```
+
+### Plan how to build SRPMs in Copr in Packit Service
+
+- create CLI command or only Packit API method for preparing sources (would mostly use existing functionality, described above)
+- implement creation of the script (Python/bash - Python would be easier for handling with configs)
+  - script will be created dynamically each time
+  - pass the needed arguments and use the method/command
+- implement using the script for creating SRPMS for Copr builds
```
TomasTomecek, line 95: would it make sense adding it inside packit?

building-srpm-in-copr/README.md
praiskup: I think we are actually talking about requirements to build _source RPMs_ not _binary RPMs_.

lbarcziova: > I think we are actually talking about requirements to build source RPMs not binary RPMs.

yes, I can make it more clear in the document

building-srpm-in-copr/README.md
lbarcziova: not sure about this, I think it makes more sense to have it only in p-s (I am thinking whether there could be any usage only in Packit), but will look into the code and rethink it

praiskup: Packit tool will be installed anyways, or?  From my point of view, after debugging similar stuff countless times, having it in some tool is always better...   there's nothing worse than logic hard-wired to a service without an easy way to reproduce the build process locally (long live MBS).

building-srpm-in-copr/README.md
```diff
@@ -90,3 +85,27 @@ copr_client.build_proxy.create_from_custom(ownername="lbarczio", projectname="og
                            script_builddeps=["git", "packit", "python3-wheel", "python3-pip", "python3-setuptools",
                             "python3-setuptools_scm", "python3-setuptools_scm_git_archive"])
 ```
+
+### Plan how to build SRPMs in Copr in Packit Service
+
+- create CLI command or only Packit API method for preparing sources (would mostly use existing functionality, described above)
+- implement creation of the script (Python/bash - Python would be easier for handling with configs)
+  - script will be created dynamically each time
```
praiskup, line 93: I touched this topic before ... but if the SRPM build method will have clear inputs (and outputs, but that's just srpm I bet), it would be relatively easy to implement a new "Packit" method in Copr ([aside](https://pagure.io/copr/copr/blob/main/f/rpmbuild/copr_rpmbuild/providers)  from the custom method).  It's up to you what you prefer, and where you want to start/iterate to.

building-srpm-in-copr/README.md
TomasTomecek: oh yeah, that's a good point, we certainly don't need *all* BRs to create an SRPM although we'll definitely need some: and we already have a list: https://github.com/packit/sandcastle/blob/main/files/install-rpm-packages.yaml#L5

PR #127, opened 2021-11-16
research `packit source-git update-source-git` command

TomasTomecek:
TODO
* [x] finish the code sync part (by the Thu meeting)
* [x] talk with the team
* [x] propose a CLI interface
* ~~[ ] create an initial implementation (if time allows)~~ we have a task for this next sprint
* [x] capture commit message: dist-git -> source-git
* [x] analysis: git-log what has changed in dist-git and check if patches or "sources" file was touched
* ~~[ ] figure out some final verification step to tell that both repositories are in sync~~ moved to a new spike card PACKIT-1828

### DIFF
--- a/update-source-git/README.md
+++ b/update-source-git/README.md
@@ -1,0 +1,109 @@
+# `packit source-git update-source-git`
+
+This research covers a command which synchronizes file changes from a dist-git
+repo back to a source-git repo.
+
+Areas to cover:
+
+- spec file synchronization
+
+- packaging sources synchronization
+
+- code synchronization
+
+Let's break these down.
+
+## Spec file sync
+
+Packit should copy the file from dist-git back to source-git. But before doing
+that, we should capture the `%version` spec file tag and check if it differs
+from the source-git spec. If it does, it means that we're dealing with a rebase
+which is not supported and packit should inform the user about this fact. See
+"Version changed" section below.
+
+If the `%release` tag has changed, packit should inform the user about the
+different release numbers. This is not an issue and is actually expected to
+happen during the proven-packager workflow when the release number is increased
+by 1 for sake of a rebuild.
+
+## Packaging sources synchronization
+
+These cover additional files in dist-git (except for patches and the "sources"
+file). Packit should detect a removal of a file and also remove it in
+source-git's `.distro` folder. Otherwise it's the same drill as with the spec
+file - copy the files.
+
+## Code synchronization
+
+There are 2 use cases in this bucket:
+
+1. Patches were changed, removed or added
+2. The package was rebased to a new version
+
+At first, we should be able to detect that such a thing has happened. There are
+multiple ways how to do that:
+
+- inspect commits and check what files have changed: if sources file or patch files
+  changed, we can easily tell that source code has changed.
+- parse spec file and compare `%version` between dist-git and source-git
+
+Packit should not support any of these 2 cases: users have to still perform
+work in their source-git repos first and not in dist-git.
+
+One problem with detecting changes is that we do not track which source-git
+commit matches a dist-git commit and vice versa. We should start with requiring
+a git-range which should be synchronized and over time figure out if this can
+be automated. Alternatively we can start putting metadata into our dist-git
+commits to tell which source-git commit they are matching.
+
+### Patches were changed
+
+1. A patch was removed
+2. A patch content changed
+3. A patch was added
+
+### Package was rebased
+
+The package was rebased to a different version which likely resulted into
+several patches being dropped and a change in git-history in source-git.
+
+## CLI proposal
+
+```
+Usage: packit source-git update-source-git [OPTIONS] DIST_GIT SOURCE_GIT REVISION_RANGE
+
+  Update a source-git repository with selected checkout of a spec file and
+  additional packaging files from a dist-git repository.
+
+  Revision range represents dist-git history which should be synchronized. Use
+  `HEAD` if you want to synchronize the current checkout. For more info how to
+  specify the revision range, see git-log(1).
+
+  If patches or the sources file changed, the command exits with return code 2.
+  Such changes are not supported by this command - code changes should always
+  happen in the source-git repo first.
+
+  This command, by default, performs only local operations and uses the
+  content of the source-git and dist-git repository as it is: does not
+  checkout branches or fetches remotes.
+
+  After the synchronization is done, packit will inform about the changes it
+  has performed and about differences between the two repositories prior to the
+  synchronization process.
+
+  Commit messages are preserved by default.
+
+  Examples:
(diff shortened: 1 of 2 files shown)


update-source-git/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+- code synchronization
+
+Let's break these down.
+
+## Spec file sync
+
+This seems trivial, we can copy the file from dist-git back to source-git. But
+before we do that, we should capture the version field and check if it differs
+from the source-git spec.
```
csomh, line 22: Should the release field also be considered?

What action should be taken if the version field is different?

update-source-git/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Sources synchronization
+
+Not much to do here, except that we should detect a removal of a file and also
+remove it in source-git's `.distro` folder.
+
+## `.distro/source-git.yaml` update
+
+Unless the version changes, there should be nothing to update. Actually this
```
csomh, line 31: What should happen to this file if the version changes?

update-source-git/README.md
```diff
@@ -0,0 +1,105 @@
...
+file should always be changed first in the source-git and there should be no
+changes happening here when getting new content from dist-git.
+
+## Code synchronization
+
+There are multiple cases in this bucket, let's talk about them individually.
+
+Cases 2) and 3) require a rebase in source-git while 1) can be fast-forwarded.
```
csomh, line 39: It's not clear what the numbers in this line refer to.

update-source-git/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Version changed
+
+The package was rebased to a different version which resulted into several
+patches being dropped.
+
+## CLI proposal
+
+```
+Usage: packit source-git update-source-git [OPTIONS] <REVISION RANGE> SOURCE_GIT DIST_GIT
```
csomh, line 76: I think it would be more natural if the order of the arguments would match the direction in which data flows. This command will take data from dist-git and put it to source-git, so the order should be:

```suggestion
Usage: packit source-git update-source-git [OPTIONS] REVISION RANGE DIST_GIT SOURCE_GIT
```

update-source-git/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+  Update a source-git repository with selected checkout of a spec file and
+  additional packaging files from the dist-git repository.
+
+  Revision range represents dist-git history which should be synchronized. Use
+  `HEAD` if you want to synchronize the current checkout.
+
+  If patches cannot be pushed in a fast-forward way to the source-git
+  repository, the command exits with return code 2. This means that only newly
+  added patches are synchronized.
```
csomh, line 86: I'm thinking that maybe we should give up on implementing synchronizing patches for good, and just not support it.

My reasoning is, that changing patch files will always come from _actual packaging work_, not a distro-wide change done by a proven packager on the spec-file or some other distro "control file" (like gating.yaml) . But if it's decided for a package to do packaging work in source-git, then they should stick to it and don't do packaging work in dist-git. We could have bots communicating this on PRs opened in dist-git by contributors who are not aware if the decision.

update-source-git/README.md
```diff
@@ -0,0 +1,105 @@
...
+    TBD
+
+Options:
+  TBD
+  -h, --help           Show this message and exit.
+```
+
+Notes:
+
+- arguments order to be consistent with update-dist-git: does that make sense?
```
csomh, line 105: See the comment above.

csomh:
> * figure out some final verification step to tell that both repositories are in sync

Would it make sense to make this check to be outside of the scope of this command and just expect that whoever calls this command has done the check and set the commit range accordingly?

update-source-git/README.md
TomasTomecek: what do you mean by considered?

I'll add a note about the version change.

update-source-git/README.md
TomasTomecek: I'll add those at the beginning of this section.

update-source-git/README.md
csomh: I mean: should there be some action taken if the release fields in dist-git and source-git are different?

TomasTomecek:
> > * figure out some final verification step to tell that both repositories are in sync
> 
> Would it make sense to make this check to be outside of the scope of this command and just expect that whoever calls this command has done the check and set the commit range accordingly?

Sadly I wasn't able to come up with an efficient way how to do that, aside from recreating the respective source-git repo and comparing them. I'll create a follow-up card.

update-source-git/README.md
TomasTomecek: Agreed, good point. I'm all for scoping the sync commands down and only support the proven packager workflow which is usually about rebuilds only.

update-source-git/README.md
TomasTomecek: okay, thanks for the explanation, I'll add a note about it

update-source-git/README.md
TomasTomecek: I just realized we don't put source-git.yaml into the dist-git repo :man_facepalming:

TomasTomecek: I added multiple clarifications about the version change in multiple sections

update-source-git/README.md
TomasTomecek: Logically that makes sense to me and I'm going to make that change (swap). My point originally was to be consistent in terms of the order of the arguments with the other update command. The reason was that I dislike in e.g. git that command arguments are very inconsistent between multiple commands.

One more change I'm going to do will be to move the `<REVISION RANGE>` to the end for two reasons:
1. it makes sense to run the command multiple times for different refs - the repo locations don't move, refs do
2. if we ever come up with a default for the range (e.g. default to HEAD), we can make it optional

update-source-git/README.md
```diff
@@ -0,0 +1,107 @@
+`packit source-git update-source-git`
```
csomh, line 1: Don't you want to make this a heading?

```suggestion
# `packit source-git update-source-git`
```

csomh (commented):
I assume the steps this command will make are:
- Take each commit in the revision range from the oldest to the newest.
- Check each commit, to make sure there were no changes to patches or the source-code.
- If the above check passed, translate each commit from dist-git to the `.distro` directory of the source-git repo (use the same commit message as in dist-git).
- If the check fails for any of the commits in the revision range, abort the whole operation and restore the content and history of the repositories. Or: we could have an option to not to do that, but instead allow a human to intervene and continue the sync from where it was interrupted (I'm really not sure if this would be useful, though, we should discuss this).

Is the above more or less correct?

csomh (approved):

PR #130, opened 2021-12-01
add "Fedora spec files analysis" research

TomasTomecek:
TODO
* [x] finish update frequency research
* [x] finish README of the research

[Preview](https://github.com/packit/research/blob/ae99e3b3d5cf27d9d77bddf0aefc81ae36f03379/fedora-spec-files/fedora-spec-files.ipynb)


fedora-spec-files/Containerfile
```diff
@@ -0,0 +1,16 @@
+FROM fedora:35
```
csomh, line 1:
```suggestion
FROM fedora:latest
```

Should be safe to go with `latest`

fedora-spec-files/Containerfile
```diff
@@ -0,0 +1,16 @@
...
+    python3-matplotlib \
+    python3-bodhi-client \
+    python3-koji \
+    python3-tabulate \
+    xz \
+    --setopt=install_weak_deps=False
+
+RUN cd / \
+    && curl -sO https://src.fedoraproject.org/repo/rpm-specs-latest.tar.xz \
+    && tar -xf ./rpm-specs-latest.tar.xz
```
csomh, line 16: For more flexibility in working with the data, I would make this a `download-data` target in the Makefile, something like:

```
rpm-specs: rpm-specs-latest.tar.xz
...

unpack-archive: rpm-specs
...

download-data: unpack-archive
...

generate-data: download-data
...
```

fedora-spec-files/generate-data.py
```diff
@@ -0,0 +1,69 @@
...
+import json
+import pathlib
+import re
+from typing import Dict, Tuple
+from bodhi.client.bindings import BodhiClient
+
+# rpm-specs-latest.tar.xz unpacked in /
+RPM_SPECS = pathlib.Path("/rpm-specs")
+# process this many packages ordered by the amount of patches
+PROCESS_PACKAGE_COUNT = 50
```
csomh, line 12: Could this be moved into the notebook, so that data for all packages is generated, but it's possible to control from the notebook how much of this data to visualize?

fedora-spec-files/generate-data.py
```diff
@@ -0,0 +1,69 @@
...
+
+    return (
+        "%autosetup" in spec_content,
+        "%setup" in spec_content,
+        "%patch" in spec_content,
+        "%autopatch" in spec_content,
+    )
+
+
+print("Generating data, please hold on...")
```
csomh, line 48: Would be nice to flush this, otherwise this is printed only when the script ends, and the user has no clue if anything is happening at all.

fedora-spec-files/Makefile
```diff
@@ -0,0 +1,18 @@
...
+# run jupyter notebook
+run:
+	@# the rooless user doesn't have access to the folder
+	@chmod -R +rwx .
+	$(PODMAN_RUN_THIS_DIR) --net=host -it $(IMAGE_NAME) jupyter notebook --allow-root ./fedora-spec-files.ipynb
+
+# Run the data collection tooling and generate data.json
+generate-data:
+	@# the rooless user doesn't have access to the folder
+	@chmod -R +rwx .
```
csomh, line 17: Things should work without this. The "rooless" user has access to all the volume mounts done with the `Z` flag.

fedora-spec-files/Makefile
```diff
@@ -0,0 +1,18 @@
...
+PODMAN_RUN_THIS_DIR := podman run --rm -v $(CURDIR):/fedora-spec-files:Z -w /fedora-spec-files
+
+# build the container image with the jupyter notebook and all the dependencies
+build:
+	podman build --tag=$(IMAGE_NAME) .
+
+# run jupyter notebook
+run:
+	@# the rooless user doesn't have access to the folder
+	@chmod -R +rwx .
```
csomh, line 11: Things should work without this. The "rooless" user has access to all the volume mounts done with the `Z` flag.

csomh (commented):
A few more comments to make this more usable in the future.

fedora-spec-files/Makefile
TomasTomecek: mine didn't, I had to chmod

fedora-spec-files/generate-data.py
TomasTomecek: pretty silly that flush is not the default

fedora-spec-files/generate-data.py
TomasTomecek: that won't work because we compute the number of builds per release from koji, which takes a few seconds for every package - doing such action for all packages would take hours

also, anyone can change that value easily, it's a simple script - let's not overcomplicate

fedora-spec-files/Makefile
csomh: Well, in order to execute `./generate-data.py` directly, the executable flag is indeed needed on it. But `u+x` on that single file and checking it in that way is enough. There is no need to make all the files in the directory executable and writable by anyone. Alternatively, you could run the script [through `python3`](https://github.com/csomh/research/commit/e63be32f3b341d881db1167b5c117ced197923c8).

fedora-spec-files/Containerfile
TomasTomecek: I'm sorry but I am not doing that - I don't know what f36 and f37 will look like: they may have completely different package names, dependencies, APIs

fedora-spec-files/Containerfile
TomasTomecek: oh, we can actually use make the way it's meant to be used, take care of files and their dependencies, this is a great suggestion

fedora-spec-files/Makefile
TomasTomecek: you're right! the perm problem was about the script not having the 'x' bit; I wish "Permission denied" was more descriptive

fedora-spec-files/Containerfile
csomh: Given that this is not a production image, I don't think this matters that much: if something brakes, the next user can fix it. But if nothing changes, they can just use this script without any change to the image.

A pinned-down OS version on the other hand can go unnoticed for longer and will certainly need some maintenance work to update it. So for me this boils down to "avoid an issue that might happen" vs "avoid the image getting outdated for sure".

TomasTomecek: I'll change it to :latest, it's not worth for me to argue about this

fedora-spec-files/generate-data.py
csomh: The :-1: meant that it's unfortunate that it's not possible to change this and make analyzing the data more flexible.

fedora-spec-files/Containerfile
csomh: As you wish.

csomh (approved):
:+1: :pray: :rocket:

fedora-spec-files/generate-data.py
TomasTomecek: roger, thanks for explanation

PR #138, opened 2022-02-25
research k8s operators

TomasTomecek:


jpopelka (commented):
Thanks!

The comment-on-issue when a fix/implementation gets deployed is a nice example but doesn't look like something we desperately need.

I haven't read about operators much, but I don't think CD is their primary use-case.
For me, there are tools focusing primarily on CD, like the ArgoCD.

But let's definitely discuss on arch.

TomasTomecek:
I'm no longer planning to finish this, especially since @jpopelka started looking into operators too.

Feel free to pick up where I left.

PR #141, opened 2022-04-19
Provide improvements for Packit service processing

lbarcziova:
Related to packit/packit-service#1364

### DIFF
--- a/improve-service-processing/README.md
+++ b/improve-service-processing/README.md
@@ -1,0 +1,101 @@
+# Improve Packit Service event processing
+
+## Problematic parts of the process
+
+- setting GH statuses is the slowest part of the short running tasks
+
+  - this shows mainly in projects with a lot of targets configured):
+
+  ```
+  [2022-04-12 21:44:01,874: DEBUG/ForkPoolWorker-1] Status reporter will report for GithubProject(namespace="osbuild", repo="osbuild"), commit=937afa51b19f5c78b552e77694d80845e69f2a84, pr=1003
+  [2022-04-12 21:44:01,874: DEBUG/ForkPoolWorker-1] Setting Github status check 'in_progress' for check 'rpm-build:centos-stream-9-x86_64': SRPM build is in progress...
+  [2022-04-12 21:44:02,553: DEBUG/ForkPoolWorker-1] Setting Github status check 'in_progress' for check 'rpm-build:fedora-36-ppc64le': SRPM build is in progress...
+  [2022-04-12 21:44:02,900: DEBUG/ForkPoolWorker-1] Setting Github status check 'in_progress' for check 'rpm-build:fedora-34-ppc64le': SRPM build is in progress...
+  [2022-04-12 21:44:03,255: DEBUG/ForkPoolWorker-1] Setting Github status check 'in_progress' for check 'rpm-build:centos-stream-8-x86_64': SRPM build is in progress...
+  [2022-04-12 21:44:03,631: DEBUG/ForkPoolWorker-1] Setting Github status check 'in_progress' for check 'rpm-build:fedora-35-ppc64le': SRPM build is in progress...
+  [2022-04-12 21:44:03,977: DEBUG/ForkPoolWorker-1] Setting Github status check 'in_progress' for check 'rpm-build:epel-8-aarch64': SRPM build is in progress...
+  [2022-04-12 21:44:04,371: DEBUG/ForkPoolWorker-1] Setting Github status check 'in_progress' for check 'rpm-build:fedora-36-aarch64': SRPM build is in progress...
+  [2022-04-12 21:44:04,698: DEBUG/ForkPoolWorker-1] Setting Github status check 'in_progress' for check 'rpm-build:fedora-36-s390x': SRPM build is in progress...
+  [2022-04-12 21:44:05,064: DEBUG/ForkPoolWorker-1] Setting Github status check 'in_progress' for check 'rpm-build:fedora-34-aarch64': SRPM build is in progress...
+  [2022-04-12 21:44:05,431: DEBUG/ForkPoolWorker-1] Setting Github status check 'in_progress' for check 'rpm-build:fedora-35-x86_64': SRPM build is in progress...
+  [2022-04-12 21:44:06,107: DEBUG/ForkPoolWorker-1] Setting Github status check 'in_progress' for check 'rpm-build:centos-stream-8-aarch64': SRPM build is in progress...
+  [2022-04-12 21:44:06,445: DEBUG/ForkPoolWorker-1] Setting Github status check 'in_progress' for check 'rpm-build:fedora-34-x86_64': SRPM build is in progress...
+  [2022-04-12 21:44:06,794: DEBUG/ForkPoolWorker-1] Setting Github status check 'in_progress' for check 'rpm-build:fedora-rawhide-ppc64le': SRPM build is in progress...
+  [2022-04-12 21:44:07,169: DEBUG/ForkPoolWorker-1] Setting Github status check 'in_progress' for check 'rpm-build:fedora-35-s390x': SRPM build is in progress...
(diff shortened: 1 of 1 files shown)


improve-service-processing/README.md
```diff
@@ -0,0 +1,73 @@
...
+  [2022-04-12 21:44:07,870: DEBUG/ForkPoolWorker-1] Setting Github status check 'in_progress' for check 'rpm-build:fedora-35-aarch64': SRPM build is in progress...
+  [2022-04-12 21:44:08,215: DEBUG/ForkPoolWorker-1] Setting Github status check 'in_progress' for check 'rpm-build:fedora-36-x86_64': SRPM build is in progress...
+  [2022-04-12 21:44:08,541: DEBUG/ForkPoolWorker-1] Setting Github status check 'in_progress' for check 'rpm-build:centos-stream-9-aarch64': SRPM build is in progress...
+  [2022-04-12 21:44:08,893: DEBUG/ForkPoolWorker-1] Setting Github status check 'in_progress' for check 'rpm-build:fedora-34-s390x': SRPM build is in progress...
+  [2022-04-12 21:44:09,227: DEBUG/ForkPoolWorker-1] Setting Github status check 'in_progress' for check 'rpm-build:epel-8-x86_64': SRPM build is in progress...
+  [2022-04-12 21:44:09,576: DEBUG/ForkPoolWorker-1] Setting Github status check 'in_progress' for check 'rpm-build:fedora-rawhide-aarch64': SRPM build is in progress...
+  ```
+
+  - we probably can't do much about it
```
lachmanfrantisek, line 34: Thankfully, we have a short-running task worker exactly for this that can be scaled up. (Maybe, we can make it slimmer so the scaling is cheaper.)

improve-service-processing/README.md
```diff
@@ -0,0 +1,73 @@
...
+## Suggestions for improvements
+
+#### Can we uninstall a repo? - Not receive events, after some period of time when we didn’t produce any meaningful work.
+
+- the documentation provides info on how to list installations, get a specific installation, delete, suspend it
+  (when a GitHub App is suspended, the app's access to the GitHub API or webhook events is blocked for that account)
+  https://docs.github.com/en/rest/reference/apps
+- we could have a periodic job which would check the installations and if an installed repo does not contain a config file
+  after some time limit (e.g. a month after installation), uninstall it
+  - if the app is installed on a namespace - at least one repo with a config file?
```
lachmanfrantisek, line 51: :+1:

improve-service-processing/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+#### Add higher prio to tasks from repos who are active users
+
+- in service we could get the repo name and check it in DB and accordingly route the tasks to different queues
+  (as we do currently for long and short running tasks)
+- the info about "active" users would be updated in the worker tasks
+  - do we want to have it only binary?
+    1. run at least one job for the repo
+    2. no jobs run
+  - or something more granular?
```
lachmanfrantisek, line 61: Maybe. Is there any built-in support for priorities in celery?

It would be nice to be able to set it runtime. I can imagine various rules (and some a bit questionable) that can potentially be combined:
* Avoid overloading service by one project. (Decrease priority if we have at least X jobs already running for this namespace/project/PR.)
* Historically active users. (Increase priority if a project has at least X jobs during the last week/month/...)
* Decrease priority of notification events (we already have two queues for this).
* Decrease priority for some triggers -- e.g. Decrease priority if the trigger is push/release.
* Increase/Decrease priority for some projects -- e.g. set it in the service-config. (Test projects can have lower priority, some privileged projects higher.)

But having too many rules can lead to issues with starvation so we don't need to overthink this. At least from the start.

improve-service-processing/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+#### Prefiltering in fedmsg
+
+- we already do prefiltering - e.g.checking Packit user for Copr builds
```
lachmanfrantisek, line 65: A related question -- Is it possible to support non-packit downstream Koji builds? (So the filtering will not be applied.)

improve-service-processing/README.md
lbarcziova: sure, I meant that we cannot fasten the process of creating a GitHub check, I will reword it

lachmanfrantisek: Ah, yes. You are right.

improve-service-processing/README.md
```diff
@@ -0,0 +1,73 @@
...
+  [2022-04-12 21:44:07,538: DEBUG/ForkPoolWorker-1] Setting Github status check 'in_progress' for check 'rpm-build:fedora-rawhide-x86_64': SRPM build is in progress...
+  [2022-04-12 21:44:07,870: DEBUG/ForkPoolWorker-1] Setting Github status check 'in_progress' for check 'rpm-build:fedora-35-aarch64': SRPM build is in progress...
+  [2022-04-12 21:44:08,215: DEBUG/ForkPoolWorker-1] Setting Github status check 'in_progress' for check 'rpm-build:fedora-36-x86_64': SRPM build is in progress...
+  [2022-04-12 21:44:08,541: DEBUG/ForkPoolWorker-1] Setting Github status check 'in_progress' for check 'rpm-build:centos-stream-9-aarch64': SRPM build is in progress...
+  [2022-04-12 21:44:08,893: DEBUG/ForkPoolWorker-1] Setting Github status check 'in_progress' for check 'rpm-build:fedora-34-s390x': SRPM build is in progress...
+  [2022-04-12 21:44:09,227: DEBUG/ForkPoolWorker-1] Setting Github status check 'in_progress' for check 'rpm-build:epel-8-x86_64': SRPM build is in progress...
+  [2022-04-12 21:44:09,576: DEBUG/ForkPoolWorker-1] Setting Github status check 'in_progress' for check 'rpm-build:fedora-rawhide-aarch64': SRPM build is in progress...
```
TomasTomecek, line 31: wow: I wonder if github doesn't provide an API to do it in a single call - we could also change the SLI to measure when the first status is updated - not sure if we're already doing it like this

improve-service-processing/README.md
```diff
@@ -0,0 +1,73 @@
...
+  ```
+
+  - we probably can't do much about it
+
+- processing event when there is no config file
+  - finding out there is no config file takes few seconds (on average around 2, but this depends probably on the repo size,
```
TomasTomecek, line 38: I think we should drop `[.]packit.json` - I haven't seen anyone using it

improve-service-processing/README.md
lbarcziova: great suggestions, I added some into the document and also added info about Celery priorities

improve-service-processing/README.md
TomasTomecek: agreed

since the workflow to enable packit again is quite clear: install it :)

improve-service-processing/README.md
lbarcziova: not sure I understand the question, you mean moving the filtering to the workers?

improve-service-processing/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+- we already do prefiltering - e.g.checking Packit user for Copr builds
+- can we do something more here?
+
+#### Recording metrics about tasks
+
+- we should first start recording more tasks-related metrics to get more insights
+  - task duration
+  - the rate of tasks processing repos without a config file
+  - ?
```
TomasTomecek, line 92: YES please

improve-service-processing/README.md
lachmanfrantisek: @lbarcziova Currently, we filter out builds not made by our FAS user (in the listener), but if we want to accept also builds for other users we could need to remove that filtering. And that goes against this... So the question is if we can actually do that without being overwhelmed by the messages...

improve-service-processing/README.md
lachmanfrantisek: :+1: Would be nice to measure this so we know for sure.

improve-service-processing/README.md
lbarcziova: I think that with the current infrastructure we don't have that strict limits on resources, so we can scale the short running workers if needed and we could also have a dedicated worker for these if it would be too overwhelming or slow down the processing of the upstream messages too much

PR #142, opened 2022-04-29
Explore more monorepo use-cases

csomh:
The original monorepo research was made for source-git. Extend this by
looking at how this can be used in upstream projects with Packit.

### DIFF
--- a/source-git-monorepos/README.md
+++ b/source-git-monorepos/README.md
@@ -14,21 +14,32 @@ To change this, a mechanism to specify which path(s) in the monorepo to map to
 which dist-git repo in the distribution needs to be introduced.
 
 This could be done by introducing a `packages` key, to hold a dictionary of
-`{package_name: package_object}`. Package-objects whould have keys identical
+`{package_name: package_object}`. Package-objects would have keys identical
 to the current top-level keys in `.packit.yaml` and `source-git.yaml`, except
 `upstream_project_url` and `upstream_ref`—these are left to be top-level keys
-only, and a new `package_paths` key which is a list of paths in the monorepo
-which map to the dist-git repo specified by `downstream_package_name`. If
-`downstream_package_name` is not specified, it would be assumed that the
+only.
+
+A package-object could also have a new `package_paths` key which is a list of
+paths in the monorepo which map to the dist-git repo specified by
+`downstream_package_name`.
+
+If `downstream_package_name` is not specified, it would be assumed that the
 dist-git repo is called `package_name`.
 
+If `package_paths` is not defined, it defaults to the path matching
+`package_name`, that is to `package_paths: [<package_name>]`.
+
 When processing a package from a monorepo, Packit should operate within these
-paths, that is, limit patch generation from Git-history to these paths.
+paths, that is, limit patch generation from Git-history to these paths, and
+use only these paths to create the source-tarballs.
 
 <!-- markdown-toc start - Don't edit this section. Run M-x markdown-toc-refresh-toc -->
 
 **Table of Contents**
 
+- [`package_paths` and subdirectories to react on](#package_paths-and-subdirectories-to-react-on)
+- [`package_paths` and working directories](#package_paths-and-working-directories)
+- [Packages and specfile discovery](#packages-and-specfile-discovery)
 - [Impact on Source-git](#impact-on-source-git)
   - [`packit source-git init`](#packit-source-git-init)
   - [`packit source-git update-dist-git`](#packit-source-git-update-dist-git)
@@ -56,6 +67,75 @@ paths, that is, limit patch generation from Git-history to these paths.
 - [Open Issues](#open-issues)
 
 <!-- markdown-toc end -->
+
+## `package_paths` and subdirectories to react on
+
+In the context of monorepos it becomes handy to be able to filter the events a
+service or bot reacts to based in which paths of the monorepo the change
+happened.
+
+This is different from `package_paths` in the sense that it allows bots and
+services to react to a broader (or different) list of paths than the ones
+defined to belong to a package.
+
+To support this a new `handle` configuration key should be introduced with a
+`path` sub-key.
+
+`handle.paths` can be defined on the top-level, in the package-level (although
+it's not going to be used by the CLI) and on the job-level, and it's inherited
+as other configuration keys.
+
+`handle.paths` extends the trigger config option of jobs, by limiting the
+validity of the trigger to the paths defined.
+
+When the `packages` key is used, jobs have their `handle.paths` set to the
+list of `package_paths` of the packages to be handled by the job. If
+`packages` is not used, the default value is the path of the root of the Git
+repo.
+
+We could mimic the GitHub syntax here, and also provide `handle.ignore_paths`
+and the exclamation-mark syntax (`handle.paths: [!exluded/path, included/path]`).
+
+Providing `handle.ignore_paths` would be interesting, b/c it could be used to
+ignore certain paths for all the packages, by defining its values on the
+top-level.
+
+Note, that a top-level `handle.paths` would be always overrode by a
+`package.paths`. A `!flatten` tag would come handy here, too, to allow to use
+references to combine lists in YAML.
+
+Introducing `handle.paths` would be independent of introducing `packages`
+(monorepo support), as it can be useful even without that, in fact [it was
+requested] some long time ago.
+
+For an example see the [Impact on Packit](#impact-on-packit) chapter below.
+
+## `package_paths` and working directories
+
+It's still not clear if having a `working_dir` configuration value to set a
+custom working directory would be strictly required for monorepo support.
+User-defined actions can always `cd` to the sub-directory of their liking. As
+for default actions, it's less complicated if they all run from the root of
+the Git repository.
+
+## Packages and specfile discovery
+
+Without `packages`:
+
+1. `specfile_path`, if defined.
+2. `<downstream_package_name>.spec`, if `downstream_package_name` set.
+3. Recursively search the Git repository.
+
+With `packages`, for each package:
+
+1. `specfile_path`, if defined.
+2. `<downstream_package_name>.spec` in all the `paths`, either because
+   `downstream_package_name` is defined for the package, or because it
+   defaults to the name of the package.
+3. Recursively search all the `paths` of the package.
+
+Both in the case of 2. and 3. `paths` are searched in the order in which they
+are defined in the configuration file.
 
 ## Impact on Source-git
 
@@ -237,9 +317,13 @@ all packages of a monorepo should be handled.
 Example:
 
 ```yaml
+handle:
+  ignore_paths:
+    - doc
 upstream_project_url: https://pagure.io/copr/copr.git
 packages:
   copr-backend:
+    # defining 'specfile_path' would be optional
     specfile_path: backend/copr-backend.spec
     package_paths:
       - backend
@@ -260,12 +344,21 @@ jobs:
     packages:
       - copr-frontend
       - copr-backend
+    handle:
+      paths:
+        - backend
+        - frontend
+        - common
   - job: copr_build
     trigger: pull_request
     targets:
       - fedora-all
     packages:
       - copr-cli
+    handle:
+      paths:
+        - cli
+        - common
 ```
 
 ### packit init
@@ -454,3 +547,4 @@ source-git configuration is read as currently is.
 [released as a batch]: https://bodhi.fedoraproject.org/updates/FEDORA-2022-8f02fc9461
 [packit/packit#88]: https://github.com/packit/packit/issues/88
 [the `metadata` key was dropped]: https://github.com/packit/packit/issues/1515
+[it was requested]: https://github.com/packit/packit-service/issues/545
--- a/source-git-monorepos/README.md
+++ b/source-git-monorepos/README.md
@@ -19,15 +19,15 @@ to the current top-level keys in `.packit.yaml` and `source-git.yaml`, except
 `upstream_project_url` and `upstream_ref`—these are left to be top-level keys
 only.
 
-A package-object could also have a new `package_paths` key which is a list of
+A package-object could also have a new `paths` key which is a list of
 paths in the monorepo which map to the dist-git repo specified by
 `downstream_package_name`.
 
 If `downstream_package_name` is not specified, it would be assumed that the
 dist-git repo is called `package_name`.
 
-If `package_paths` is not defined, it defaults to the path matching
-`package_name`, that is to `package_paths: [<package_name>]`.
+If `package.paths` is not defined, it defaults to the path matching
+`package_name`, that is to `paths: [<package_name>]`.
 
 When processing a package from a monorepo, Packit should operate within these
 paths, that is, limit patch generation from Git-history to these paths, and
@@ -37,8 +37,8 @@ use only these paths to create the source-tarballs.
 
 **Table of Contents**
 
-- [`package_paths` and subdirectories to react on](#package_paths-and-subdirectories-to-react-on)
-- [`package_paths` and working directories](#package_paths-and-working-directories)
+- [`package.paths` and subdirectories to react on](#packagepaths-and-subdirectories-to-react-on)
+- [`package.paths` and working directories](#packagepaths-and-working-directories)
 - [Packages and specfile discovery](#packages-and-specfile-discovery)
 - [Impact on Source-git](#impact-on-source-git)
   - [`packit source-git init`](#packit-source-git-init)
@@ -68,13 +68,13 @@ use only these paths to create the source-tarballs.
 
 <!-- markdown-toc end -->
 
-## `package_paths` and subdirectories to react on
+## `package.paths` and subdirectories to react on
 
 In the context of monorepos it becomes handy to be able to filter the events a
 service or bot reacts to based in which paths of the monorepo the change
 happened.
 
-This is different from `package_paths` in the sense that it allows bots and
+This is different from `package.paths` in the sense that it allows bots and
 services to react to a broader (or different) list of paths than the ones
 defined to belong to a package.
 
@@ -89,7 +89,7 @@ as other configuration keys.
 validity of the trigger to the paths defined.
 
 When the `packages` key is used, jobs have their `handle.paths` set to the
-list of `package_paths` of the packages to be handled by the job. If
+list of `package.paths` of the packages to be handled by the job. If
 `packages` is not used, the default value is the path of the root of the Git
 repo.
 
@@ -110,7 +110,7 @@ requested] some long time ago.
 
 For an example see the [Impact on Packit](#impact-on-packit) chapter below.
 
-## `package_paths` and working directories
+## `package.paths` and working directories
 
 It's still not clear if having a `working_dir` configuration value to set a
 custom working directory would be strictly required for monorepo support.
@@ -148,7 +148,7 @@ packages:
   lit:
     downstream_package_name: python-lit
     specfile_path: .distro/python-lit/python-lit.spec
-    package_paths:
+    paths:
       - llvm/utils/lit
     # This could be also used, but it's not required here.
     # patch_generation_ignore_paths:
@@ -209,15 +209,15 @@ Example `packages.yaml`:
 
 ```yaml
 - downstream_package_name: copr-backend
-  package_path: backend
+  path: backend
   dist_git_path: rpms/copr-backend
   dist_git_branch: rawhide
 - downstream_package_name: copr-frontend
-  package_path: frontend
+  path: frontend
   dist_git_path: rpms/copr-frontend
   dist_git_branch: rawhide
 - downstream_package_name: copr-cli
-  package_path: cli
+  path: cli
   dist_git_path: rpms/copr-cli
   dist_git_branch: rawhide
 ```
@@ -295,7 +295,7 @@ match `basename DIST_GIT`.
 
 When opening an MR in a source-git monorepo, Hardly should figure out which
 dist-git packages are going to be impacted by the change (by looking at
-`package_paths`), and open mirror-MRs in the dist-git repositories of those
+`package.paths`), and open mirror-MRs in the dist-git repositories of those
 packages.
 
 CI results from all mirror-MRs should be synced back to the source-git MR.
@@ -325,15 +325,15 @@ packages:
   copr-backend:
     # defining 'specfile_path' would be optional
     specfile_path: backend/copr-backend.spec
-    package_paths:
+    paths:
       - backend
   copr-frontend:
     specfile_path: frontend/copr-frontend.spec
-    package_paths:
+    paths:
       - frontend
   copr-cli:
     specfile_path: cli/copr-cli.spec
-    package_paths:
+    paths:
       - cli
 jobs:
   - job: copr_build
@@ -433,7 +433,7 @@ The command should display status for all packages in a monorepo.
 
 ### packit validate-config
 
-Check if the paths specified in `package_paths` do exist in the monorepo.
+Check if the paths specified in `package.paths` do exist in the monorepo.
 
 Check if the package names (`downstream_package_name`) are correct and indeed
 exist in the distribution. See [packit/packit#88].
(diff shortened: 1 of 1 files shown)


csomh:
Okay, I just realized, that maybe `packages` should be a dictionary instead of a list, as we would expect that each item in `packages` is a different package, with a different name.

If the schema is changed like this, then the package name could be used as a default for `downstream_package_name`.

Examples:

```yaml
packages:
  backend:
    downstream_package_name: copr-backend
    ...
  frontend:
    downstream_package_name: copr-frontend
    ...

# When referencing a package, package name (the key from the 'packages' dict) should be used
jobs:
- job: copr_build
  trigger: pull_request
  packages:
  - backend
  - frontend
```

The above could be written without `downstream_package_name`:

```yaml
packages:
  copr-backend:
    ...
  copr-frontend:
    ...

# When referencing a package, package name (the key from the 'packages' dict) should be used
jobs:
- job: copr_build
  trigger: pull_request
  packages:
  - copr-backend
  - copr-frontend
```

@lachmanfrantisek @TomasTomecek WDYT? Should I change the proposal like this?

jpopelka (commented):
looks OK so far, +1 to change `packages` to a dictionary

source-git-monorepos/README.md
```diff
@@ -1,41 +1,80 @@
-# Support monorepos in source-git
+# Support for Monorepos
 
-Monorepos store the source code of multiple projects in a single
-repository. They are a special use-case from a source-git point of view,
-because an upstream monorepo maps to multiple dist-git repositories in the
-distribution.
+Monorepos store the source code of multiple projects in a single repository.
```
TomasTomecek, line 3: I wonder if want to account for different build configurations too, e.g. have different compiler flags for multiple sets of builds.

source-git-monorepos/README.md
```diff
@@ -50,21 +89,355 @@ packages:
...
+kept in subdirectories, named according to `package_name`, in `.distro`, to
 keep things organized.
 
 This way all distribution configs and files would be kept at one place, which
-probably is somewhat better than having source-git configuration spread
-across the monorepo, in subdirectories. This setup also allows multiple paths
-from the repo mapped to a dist-git repo.
+probably is somewhat better than having source-git configuration spread across
+the monorepo, in subdirectories. This setup also allows multiple paths from
+the repo mapped to a dist-git repo.
```
TomasTomecek, line 98: I'm confused here: so where would I find a specfile?

source-git-monorepos/README.md
```diff
@@ -50,21 +89,355 @@ packages:
...
+(`PackitAPI.up`) to multiple dist-git/downstream repositories. This means that
+`PackitAPI.dg` should become a dictionary holding the objects for each
+dist-git repository.
+
+In order to make the transition easier, we could keep `.dg` as it is, and make
+it to return the dist-git repo for the first (and only) package. If there are
+more packages defined, an exception should be raised. This way parts of the
+API and CLI which were not yet updated to support monorepos, could continue to
+work, but would fail if used with a monorepo configuration specifying multiple
+packages.
```
TomasTomecek, line 355: Oh my, this will indeed be a lot of work

TomasTomecek (commented):
This looks like a great hackfest material.

source-git-monorepos/README.md
csomh: I would expect to find things related to build configuration in the spec-file of each package (or some other scripting called during *build-time*), and from this point of view I would say these fall outside the scope of source-git. Or am I missing something?

source-git-monorepos/README.md
csomh: By convention in source-git specfiles would be stored in `.distro/<package_name>/<dist_git_name>.spec`. I've added an example above to make this more clear.

In a Packit context, it doesn't really matter, as everything is configurable.

TomasTomecek: That makes sense, I like that the spec files are in a single place. I thought about how archives for these packages are created - but Packit can already work with setups where spec files and sources are in different directories.

source-git-monorepos/README.md
TomasTomecek: I am thinking out loud here as I see a big potential in the monorepo concept that it could solve more use cases related to building.

rpmbuild and mock [support](https://github.com/rpm-software-management/mock/blob/main/mock/docs/mock.1#L417) setting [conditionals build options](https://rpm-software-management.github.io/rpm/manual/conditionalbuilds.html). AFAIK no one requested this functionality from us yet (not used anymore?). Nevertheless, I just wanted to mention the feature here.

No need to change anything in the research doc. Let's keep the scope down.

mfocko: I'd be more afraid about shared sources between packages, e.g. helpers, utility libraries.

_edit_: never mind, there's `package_paths`

source-git-monorepos/README.md
```diff
@@ -1,41 +1,80 @@
...
+`upstream_project_url` and `upstream_ref`—these are left to be top-level keys
+only, and a new `package_paths` key which is a list of paths in the monorepo
+which map to the dist-git repo specified by `downstream_package_name`. If
+`downstream_package_name` is not specified, it would be assumed that the
+dist-git repo is called `package_name`.
+
+When processing a package from a monorepo, Packit should operate within these
+paths, that is, limit patch generation from Git-history to these paths.
+
+<!-- markdown-toc start - Don't edit this section. Run M-x markdown-toc-refresh-toc -->
```
mfocko, line 28: (definitely can't tell the editor :eyes:)

source-git-monorepos/README.md
```diff
@@ -50,21 +89,369 @@ packages:
...
 
-For **backwards compatibility**: if the top-level `packages` key is missing,
+Check if the paths specified in `package_paths` do exist in the monorepo.
+
+Check if the package names (`downstream_package_name`) are correct and indeed
+exist in the distribution. See [packit/packit#88].
+
+### packit prepare-sources
+
+I'm not entirely sure my understanding of what this command does is correct 😅
```
mfocko, line 350: It's run as a preparation step when building SRPMs in Copr.

source-git-monorepos/README.md
```diff
@@ -50,21 +89,369 @@ packages:
...
+
+In the example above when building PRs in Copr, the `srpm_build_deps` defined
+in the job would be used for both packages.
+
+On the other hand, when building releases in Copr, only `subpackage1` is going
+to be built, and the `srpm_build_deps` will be the ones defined for the
+package in the `packages` section.
+
+The example above could be made more useful if extending lists would be
+supported by Packit. (YAML doesn't support this, one gets nested lists when
```
mfocko, line 428: fmf… but that would be another story I guess

source-git-monorepos/README.md
csomh: :see_no_evil:

source-git-monorepos/README.md
csomh: Yes, it would another story. This could be even done separately from this whole monorepo-saga. Stackowerflow suggested something like:

```yaml
list1: &listref
- element1
- element2
list2: !flatten
- *listref
- additional-element
```

We would need to write the code to flatten `[['element1', 'element2'], 'additional-element']`, and set a trigger to do it when the `!flatten` tag is met.

TomasTomecek (approved):

source-git-monorepos/README.md
```diff
@@ -50,21 +89,367 @@ packages:
...
+using references in lists.)
+
+### Impact on Actions
+
+Because actions can now be defined on the top-level, package-level and
+job-level, one might ask, in what path these should be executed.
+
+Because a package can map to multiple sub-paths in the monorepo, the most
+sensible thing to do is to keep it simple and execute each action at the
+top-level, even if the action is defined on a job- or package-level.
```
lachmanfrantisek, line 436: Yes, that was my concern in the first question. But if we preserve current use-cases and document the behaviour of monorepo users, I don't see a problem with choosing any approach.

lachmanfrantisek (commented):
Wow, that's really detailed!

source-git-monorepos/README.md
```diff
@@ -50,21 +89,367 @@ packages:
...
+As an alternative, the command could have a variant where it takes a basic
+`packages` definition to guide the initialization of a source-git monorepo.
+
+Example `packages.yaml`:
+
+```yaml
+- downstream_package_name: copr-backend
+  package_path: backend
+  dist_git_path: rpms/copr-backend
+  dist_git_branch: rawhide
```
praiskup, line 134: I tend to feel that this should be a list of branch names?

source-git-monorepos/README.md
```diff
@@ -50,21 +89,367 @@ packages:
...
+On the other hand, the command needs to be updated to support a source-git
+repo as a target when the source-git repo is a monorepo. In this case only the
+paths of the sub-package need to be considered during the update. `DIST_GIT`
+should match one of the `downstream_package_name`s in `.distro/source-git.yaml`.
+
+### `packit source-git status`
+
+The command should check the sync status of a sub-package from a source-git
+monorepo and the corresponding dist-git repo. `downstream_package_name` should
+match `basename DIST_DIS`.
```
praiskup, line 212: DIST_DIS?

source-git-monorepos/README.md
```diff
@@ -50,21 +89,367 @@ packages:
...
+  dist_git_path: rpms/copr-backend
+  dist_git_branch: rawhide
+- downstream_package_name: copr-frontend
+  package_path: frontend
+  dist_git_path: rpms/copr-frontend
+  dist_git_branch: rawhide
+- downstream_package_name: copr-cli
+  package_path: cli
+  dist_git_path: rpms/copr-cli
+  dist_git_branch: rawhide
```
praiskup, line 142: I bet only Fedora dist-git is supported, now or forever?  Copr identifies also CentOS, CentOS Stream and Red Hat (internally). Plus four Copr dist-git instances, but these are not open for PRs.  So just a feedback, DistGit operations in Copr have the `--distgit <DISTGIT_ID` option (which hides the dist-git instance configuration details).

source-git-monorepos/README.md
```diff
@@ -50,21 +89,367 @@ packages:
...
+**At a minimum**, the command should be updated to support monorepos, and
+transform only the content specific to the package from source-git. The
+package name (`downstream_package_name`) to be considered from
+`.distro/source-git.yaml` is the one which matches `basename DIST_GIT`.
 
-For **backwards compatibility**: if the top-level `packages` key is missing,
+**Going further**, when used with monorepos, this command could be able to
+update multiple dist-git repositories. This means, that the `DIST_GIT`
+argument of the command would become `DIST_GIT [DIST_GIT...]` (one or more
+paths to dist-git repositories).
```
praiskup, line 171: In Tito, you do `tito release` from within the upstream sub-directory.  E.g. you
do `cd backend && tito release <releaser_id>`.  Not sure what is better, but just
noting that it might be worth considering.

source-git-monorepos/README.md
```diff
@@ -50,21 +89,367 @@ packages:
...
+
+Let's not change the interface of this command and don't support the usecase
+to update a single source-git monorepo from multiple dist-git repositories in
+order to avoid complications. Expect that multiple commands are going to be
+executed if updates need to be taken from multiple dist-git repositories.
+
+On the other hand, the command needs to be updated to support a source-git
+repo as a target when the source-git repo is a monorepo. In this case only the
+paths of the sub-package need to be considered during the update. `DIST_GIT`
+should match one of the `downstream_package_name`s in `.distro/source-git.yaml`.
```
praiskup, line 206: This command could IMO be called also from within the source-git subdirectory.

source-git-monorepos/README.md
```diff
@@ -50,21 +89,367 @@ packages:
...
+### `packit source-git status`
+
+The command should check the sync status of a sub-package from a source-git
+monorepo and the corresponding dist-git repo. `downstream_package_name` should
+match `basename DIST_DIS`.
+
+### Hardly
+
+When opening an MR in a source-git monorepo, Hardly should figure out which
+dist-git packages are going to be impacted by the changed (by looking at
```
praiskup, line 217: s/changed/change/

source-git-monorepos/README.md
```diff
@@ -50,21 +89,367 @@ packages:
...
+
+Example:
+
+```yaml
+upstream_project_url: https://pagure.io/copr/copr.git
+packages:
+  copr-backend:
+    specfile_path: backend/copr-backend.spec
+    package_paths:
+      - backend
```
praiskup, line 245: In Copr we by default search for `<package_path>/<package_name>.spec`, to simplify the config.

source-git-monorepos/README.md
```diff
@@ -50,21 +89,367 @@ packages:
...
+
+### packit init
+
+No changes are required for this command. `.packit.yaml` is expected to be
+crafted manually if the upstream repository is a monorepo.
+
+### packit srpm
+
+The command should receive a new optional argument, `[PACKAGE...]`, to define
+zero or more packages for which an SRPM should be built.
```
praiskup, line 279: Ditto here, having a possibility to call this also from within sub-dir would be plus.

source-git-monorepos/README.md
```diff
@@ -1,41 +1,80 @@
...
 
 To change this, a mechanism to specify which path(s) in the monorepo to map to
-which dist-git repo in the distribution should be introduced.
+which dist-git repo in the distribution needs to be introduced.
+
+This could be done by introducing a `packages` key, to hold a dictionary of
+`{package_name: package_object}`. Package-objects whould have keys identical
+to the current top-level keys in `.packit.yaml` and `source-git.yaml`, except
+`upstream_project_url` and `upstream_ref`—these are left to be top-level keys
+only, and a new `package_paths` key which is a list of paths in the monorepo
```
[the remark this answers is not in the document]
csomh: > I have a few questions about `package_paths`:
> 
> * Is this the opposite of [RFE: Excluded paths packit-service#545](https://github.com/packit/packit-service/issues/545) (=react only on changes in this path(s))? Or use it just for source manipulation like archive/path generation? Or, a working directory for our jobs? Or all?
>   
>   * Maybe, we can introduce an option for each usecase -- `source_path`, `include_path`/`exclude_path`, `working_path`.

That's a good question.

I see `package_paths` as only used for archive and patch generation. The _Excluded paths_ RFE is different in the sense that it tells when to run or not to run Packit.

`package_paths` might also be used for this purpose, but that might be more problematic: packages in the monorepo might be interdependent, and some projects would like to build all of them on every change, other projects just the packages which had code changed. So I would see that as a separate config key.

As for `working_path`, I think it makes sense to keep that at the root of the repository, to not to complicate things.

> * Will this be required? (I can think of some use-cases without having distinct subdirectories for each package.)

I would be curious to see some examples for this. So far the monorepos I've seen (llvm, copr) have generated packages from one or more sub-directories in the directory.

source-git-monorepos/README.md
```diff
@@ -50,21 +89,367 @@ packages:
...
+├── frontend
+│   └── copr-frontend.spec
+└── source-git.yaml
+```
 
 This way all distribution configs and files would be kept at one place, which
-probably is somewhat better than having source-git configuration spread
-across the monorepo, in subdirectories. This setup also allows multiple paths
-from the repo mapped to a dist-git repo.
+probably is somewhat better than having source-git configuration spread across
```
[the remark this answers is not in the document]
csomh: This sentence is in the source-git section, so I'll probably not apply this suggestion :)

In source-git I think it makes sense to have a more opinionated layout of the repository in order to be consistent across packages in a distribution. For Packit, I don't know if this is important. Upstream projects using Packit might choose to put files to different places.

One thing we should avoid is to encourage projects using Packit to create a `.distro` directory upstream, as that would conflict with the source-git layout :D

source-git-monorepos/README.md
csomh: Why so?

`source-git init` initializes a single branch in a source-git repo. Furthermore: source-git branches map to dist-git branches one-to-one.

source-git-monorepos/README.md
```diff
@@ -50,21 +89,367 @@ packages:
...
+the monorepo, in subdirectories. This setup also allows multiple paths from
+the repo mapped to a dist-git repo.
```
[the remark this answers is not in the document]
csomh: Added "single" to be more explicit.

source-git-monorepos/README.md
csomh: All dist-gits are supported, I just tend to use Fedora in examples.

As of now mapping the source-git repo to a dist-git instance is not configurable for the users only for the bots, as we expect CentOS to want to have their own source-git repos in their own place (which are mapped to CentOS dist-git) and Fedora to want their own source-git repos in their own place, mapped to Fedora dist-git repos.

source-git-monorepos/README.md
csomh: Thanks for mentioning this.

source-git-monorepos/README.md
csomh: This command could be called from any path as it'll receive 3 arguments: the path to the dist-git repository, the path to the source-git repo, and a revision range in dist-git.

source-git-monorepos/README.md
csomh: Thanks!

source-git-monorepos/README.md
```diff
@@ -50,21 +89,367 @@ packages:
...
+upstream_project_url: https://pagure.io/copr/copr.git
+packages:
+  copr-backend:
+    specfile_path: backend/copr-backend.spec
+    package_paths:
+      - backend
+  copr-frontend:
+    specfile_path: frontend/copr-frontend.spec
+    package_paths:
+      - frontend
```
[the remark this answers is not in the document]
csomh: Actually it does not mean that. `package_paths` only maps monorepo paths to dist-git repositories. But it doesn't tell when actions using that mapping should be done. I think I'll need to revisit this topic once again and be more detailed about the intent, b/c @lachmanfrantisek brought up a similar topic.

source-git-monorepos/README.md
csomh: Although plain `packit` commands prefer to guess their arguments from their context, context might be ambiguous so I try to follow the "explicit is better than implicit" mantra to avoid confusion.

If the current proposal is going to be followed than calling `packit srpm` from a sub-dir would result in building an SRPM for all the packages in the monorepo (Packit would look up the config in the repo-root, and b/c there is not `PACKAGE` argument provided, it will build an SRPM for all the packages defined in the config.

Of course, we could say, that if the current sub-dir is a `package_path` for a package in the config, then build just that SRPM. Which is fine, if there is only one `package_path` in a package, but becomes more complicated if there are two or more. And to avoid this complication and possible user confusion, I would prefer to stay explicit :)

source-git-monorepos/README.md
```diff
@@ -50,21 +89,367 @@ packages:
...
+to limit changes only to select packages.
+
+## Impact on Packit API
+
+### Internal Design
+
+With monorepos needs to be taught to map a single upstream repository
+(`PackitAPI.up`) to multiple dist-git/downstream repositories. This means that
+`PackitAPI.dg` should become a dictionary holding the objects for each
+dist-git repository.
```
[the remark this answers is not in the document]
csomh: Wouldn't this approach leak logic regarding the interpretation of the configuration to the CLI or the service and potentially cause re-implementing similar logic in both places?

source-git-monorepos/README.md
praiskup: Hm, I guess? I'm offtopic (I thought it is the shared config format with the packit service, and the "propose downstream" configuration).

Just from the experience with tito... where you configure the precise set of branches you want to automatically update with new releases (aka `tito release`).

source-git-monorepos/README.md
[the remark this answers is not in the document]
praiskup: Yeah, I just feel mandatory to saying this aloud :-) This is an absolute must for monorepos like Copr is, we simply couldn't afford to rebuild all the large set packages for each PR.  On top of this, it would be nice to have a way to define "order" for building them, in case multiple of them are affected by a single PR and there are dependencies between them.

source-git-monorepos/README.md
csomh: The only example I know is in LLVM. For their daily builds, the do something source-git-like, and for `clang` they [fetch 2 paths](https://github.com/kwk/llvm-project/blob/0eacd884dd720b7a2156fb84df910388c5677b33/clang/clang.spec#L88) from their monorepo.

source-git-monorepos/README.md
```diff
@@ -50,21 +89,367 @@ packages:
...
+      - epel-8
+    packages:
+      - copr-frontend
+      - copr-backend
+  - job: copr_build
+    trigger: pull_request
+    targets:
+      - fedora-all
+    packages:
+      - copr-cli
```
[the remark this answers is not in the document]
csomh: Obviously this is just and example for how the configuration should look like, which by chance it uses `copr`, but in no way should this be considered as a reference config for `copr`.

source-git-monorepos/README.md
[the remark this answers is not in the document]
csomh: I took some of your suggestions and made a proposal to have `handle.paths`, in order to be able to fine tune what paths services should react to (or not), independent of `package.paths`. See the "`package.paths` and subdirectories to react on" chapter.

I'm still not sure about working directories, but I've took note of this uncertainty and made a separate chapter for this, too.  Of course we can discuss this further if you want: I'm open to be convinced that this is essential to have a good monorepo support.

source-git-monorepos/README.md
csomh: This configuration is a proposal that could make **initializing** source-git monorepos from dist-git repositories easier. It's not planned to be used as a source-git config per se.

source-git-monorepos/README.md
csomh: I've made a proposal for this: see the "Package and specfile discovery" chapter.

source-git-monorepos/README.md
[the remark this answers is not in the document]
csomh: Added a new chapter "`package.paths` and subdirectories to react on", which makes a proposal to deal with this.

csomh:
This PR was updated to include a proposal for `handle.paths`, clarified some defaults, renamed `package_paths` to `package.paths`, and updated examples accordingly.

csomh:
Let's keep this open until all talking points are settled with @lachmanfrantisek  and @praiskup.

source-git-monorepos/README.md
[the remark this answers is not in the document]
praiskup: Since nobody except for us will use/read this example, I think it is fine.  But if you wanted to use this as an example in docs, I would prefer switching it (clients supported everywhere, servers only on Fedora, as more (and newer) deps are typically needed).

praiskup:
> Let's keep this open until all talking points are settled with @lachmanfrantisek and @praiskup.

I'm fine here.  Thank you for the heads-up!

source-git-monorepos/README.md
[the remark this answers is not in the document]
lachmanfrantisek: > I'm open to be convinced that this is essential to have a good monorepo support.

No, that's maybe more an enhancement.

> I would be curious to see some examples for this. So far the monorepos I've seen (llvm, copr) have generated packages from one or more sub-directories in the directory.

I know it's not a true monorepo, but e.g. having multiple different spec-files using the same source tree -- e.g. `/fedora/some.spec` and `/rhel/some.spec`, but source in `/project`.

source-git-monorepos/README.md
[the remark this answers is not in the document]
lachmanfrantisek: I don't think it's a huge issue. Maybe, we can share the common code at least. This is the first approach I would try but maybe we will find some issues when working on this for real.

source-git-monorepos/README.md
lachmanfrantisek: :+1: Exactly.

source-git-monorepos/README.md
```diff
@@ -50,21 +89,367 @@ packages:
...
 source-git configuration is read as currently is.
 
+## Open Issues
+
 - TODO Can archives stored in the dist-git repositories mapped to one or more
   paths in the upstream repo?
 - TODO Is support needed to have different packages generated from different
   git refs? (This would defy the point in having a monorepo, but some might
-  require it.)
+  require it, for example [Copr].)
```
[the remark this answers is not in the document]
lachmanfrantisek: E.g. Anaconda uses a different branch and packit config for each distribution -- and there are basically two sets of releases -- one made from a Fedora branch and the other one from the RHEL branch.
In that case, you have a different config in each branch so I am now not sure if this needs to be handled anyhow.

PR #144, opened 2022-05-16
Reserach more metric types for monitoring p-s

lbarcziova:


monitoring/metrics.md
```diff
@@ -0,0 +1,163 @@
+## Requests coming to / of the API
```
TomasTomecek, line 1: since we already measure the amount of Github webhooks we receive, this would give us a more granular view into the API as a whole: and transitively into dashboard, since it's the biggest user of the API

monitoring/metrics.md
```diff
@@ -0,0 +1,163 @@
...
+    - celery_task_sent_total
+    - celery_task_received_total
+    - celery_task_started_total
+    - celery_task_succeeded_total
+    - celery_task_failed_total
+    - celery_task_rejected_total
+    - celery_task_revoked_total
+    - celery_task_retried_total
+    - celery_worker_up
+    - celery_worker_tasks_active
```
TomasTomecek, line 34: these seem mega useful! especially since we don't measure anything in fedora-messaging - we can easily get overwhelmed from it without us noticing

TomasTomecek (commented):
What do you suggest we should do next?

lbarcziova:
> What do you suggest we should do next?

I would definitely like to quickly go through these on architecture and get opinions on what should we prioritize, but I think we can start with celery metrics since that is already planned and then next the API metrics could be also useful.

TomasTomecek:
+1 celery metrics

mfocko (approved):
lgtm

PR #145, opened 2022-06-08
Adoption metrics: define what we want to measure

nikromen:
This PR exactly tells what Packit team may want to measure for their adoption metrics.

Questions to the team members:
- What metrics are you missing?
- What metrics are unnecessary in your opinion?
- What description of metrics are unclear to you?
- What metrics can be modified to have better indicative value?

I'd love to hear your views :)


adoption-metrics/define_adoption_metrics.md
```diff
@@ -0,0 +1,56 @@
+# What adoption metrics Packit team wants to measure
+
+## Static meaurement (updated periodically)
+
+### Newcomers adoption
+
+- Total amount of newcomers in a given period (monthly/quarterly/half yearly/yearly).
```
TomasTomecek, line 7: nice, I like this one - would be great to see the long-term trend in adoption

adoption-metrics/define_adoption_metrics.md
```diff
@@ -0,0 +1,56 @@
...
+- How often are different services (COPR, TF, all downstream jobs) used by active users?
+  What they use most often and what they use less?
+- How frequently active users use packit services?
+
+### Users retention
+
+- Percentage of previous newcomers who became active Packit users.
+  ((amount of newcomers from previous period becoming an active users
+  / amount of newcomers from previous period) \* 100)
+- What percentage of users are active users? (How many inactive users do we have?)
```
TomasTomecek, line 29: also a great indicator for us that those folks were not satisfied and probably moved somewhere else

adoption-metrics/define_adoption_metrics.md
```diff
@@ -0,0 +1,56 @@
...
+
+### Usage of Packit services
+
+- Percentage representation of individual services e.g. (COPR requests / all requests) \* 100
+
+## The following metrics will be displayed as a histograms
+
+As the metrics above - these will be updated once a given period.
+
+- Show most 15/20 active project of ... COPR, TF, all downstream jobs
```
TomasTomecek, line 39: +1

adoption-metrics/define_adoption_metrics.md
```diff
@@ -0,0 +1,56 @@
...
+![alt text](https://packit.dev/images/2021-in-numbers-top20-copr-builds.png)
+
+## Real time metrics
+
+This is heavily inspired by [COPR's real time stats](https://copr.fedorainfracloud.org/status/stats/)
+
+Real time graph will show overall activity of users in a given period (day/week/quarter)
+
+- Make a graph for each service packit provides (COPR, TF, all downstream jobs) and display
+  activity. This may help to see and identify peaks and lows in overall activity of users.
```
TomasTomecek, line 53: yes please!

TomasTomecek (commented):
Jirko, overall you produced a solid list of metrics, some will be very handy. I believe some would be pretty hard to implement, so I suggest picking 5-7 and focus on implementing those.

adoption-metrics/define_adoption_metrics.md
```diff
@@ -0,0 +1,56 @@
...
+- Percentage of previous newcomers who became active Packit users.
+  ((amount of newcomers from previous period becoming an active users
+  / amount of newcomers from previous period) \* 100)
+- What percentage of users are active users? (How many inactive users do we have?)
+
+### Usage of Packit services
+
+- Percentage representation of individual services e.g. (COPR requests / all requests) \* 100
+
+## The following metrics will be displayed as a histograms
```
xsuchy, line 35: I would like to see it as a heatmap similar to https://www.perfecttrendsystem.com/images/articles/finviz-stock-screener/07-Finviz-stock-heatmap.PNG

adoption-metrics/define_adoption_metrics.md
xsuchy: The real-time data in Copr are for admins to identify issues. It may be hard to set up warning levels. E.g. there are no problems if there appear 4k actions, but because actions are mostly no-op, it should go down almost immediately. If not, it may indicate a problem. Or - there is no problem if there are 40k builds in a queue, but the chart should decline. If not, it may indicate a problem in a scheduler or spawning builds.  So for real-time data, it is useful to plot data, which helps admin spot anomalies in the behaving of the whole system.

adoption-metrics/define_adoption_metrics.md
```diff
@@ -0,0 +1,56 @@
...
+
+- How often are different services (COPR, TF, all downstream jobs) used by active users?
+  What they use most often and what they use less?
+- How frequently active users use packit services?
+
+### Users retention
+
+- Percentage of previous newcomers who became active Packit users.
+  ((amount of newcomers from previous period becoming an active users
+  / amount of newcomers from previous period) \* 100)
```
lachmanfrantisek, line 28: We should somehow get rid of the projects that have the Packit installed only because of the `install GitHub application to all repositories`. We can check the presence of Packit config or the namespace to be approved. It's also important who we consider an active user.

adoption-metrics/define_adoption_metrics.md
```diff
@@ -0,0 +1,56 @@
...
+- What percentage of users are active users? (How many inactive users do we have?)
+
+### Usage of Packit services
+
+- Percentage representation of individual services e.g. (COPR requests / all requests) \* 100
```
lachmanfrantisek, line 33: Here, it will be really tricky to pick the right number for each job-- target vs group of targets vs downstream PR ...

lachmanfrantisek (commented):
Looks really nice!

The most relevant in my opinion are these:
* Number of (active) users and newcomers.
* Real-time usage.
* Top X project per job with the number of builds/runs/... (With links to their project as an inspiration.)

I would also like to see the overall number of builds/runs/...  Ideally monthly/yearly/since-the-beginning. So we can say "Packit has triggered  X Copr builds,  Y test runs and opened Z downstream PRs."

TBH, I am also impatiently checking the Packit's Fedora badges: https://badges.fedoraproject.org/user/packit ..;)

adoption-metrics/define_adoption_metrics.md
nikromen: I'll add a definition of newcomer - "Newcomers are users who have an approved namespace". 
Would it be enough?

adoption-metrics/define_adoption_metrics.md
nikromen: I was thinking of the group of targets as the basic unit of work. As we have it displayed on dashboard/jobs. Each "row" will be the unit of job. It will be easy to measure it this way (once propose-downstreams mess in DB will be fixed) and also reasonable. E.g. all the copr builds for each targets (let's say 4 targets) are triggered at once so it doesn't make sense to count it as 4 jobs but rather as one job.

adoption-metrics/define_adoption_metrics.md
nikromen: The heatmap is less clear to me than the histogram, but that's my view. What do others think?

nikromen:
I removed some metrics which are hard to implement and which requires additional logic to the code which will affect the overall performance.

Also I was thiking of one more metric: What is our average response time?
This should be fairly easy to implement. Start with measuring when we receive a message and stop measuring time when we give :eyes: reaction since this is first response from us that users see. WDYT?

It would also be useful to consider whether the metrics are sufficient to detect errors. For example, if the request queue is growing and requests are slow to process (slower than normal) etc. Can we measure these (to detect potential bugs) or are these enough?

TomasTomecek:
> Also I was thiking of one more metric: What is our average response time? This should be fairly easy to implement. Start with measuring when we receive a message and stop measuring time when we give eyes reaction since this is first response from us that users see. WDYT?

We already have a related metric for our SLO1: https://packit.dev/docs/service-level-objectives/#slo1-changes-to-github-prs-receive-a-status-update-within-15-seconds-in-99-of-cases

> It would also be useful to consider whether the metrics are sufficient to detect errors. For example, if the request queue is growing and requests are slow to process (slower than normal) etc. Can we measure these (to detect potential bugs) or are these enough?

This should be easy to do and see once we have the celery metrics, @majamassarini is working on this right now.

adoption-metrics/define_adoption_metrics.md
```diff
@@ -0,0 +1,50 @@
...
+
+### Newcomers adoption
+
+Newcomers are users who have an approved namespace.
+
+- Total amount of newcomers in a given period (monthly/quarterly/half yearly/yearly).
+
+### Active users
+
+Active users are those from whom we have recorded at least one activity in the past period
```
TomasTomecek, line 13: worth clarifying what exactly past period is :)

adoption-metrics/define_adoption_metrics.md
```diff
@@ -0,0 +1,50 @@
...
+and they do not fit into the category of newcomers.
+
+### Users retention
+
+- Total amount of active users.
+- Percentage of previous newcomers who became active Packit users.
+  ((amount of newcomers from previous period becoming an active users
+  / amount of newcomers from previous period) \* 100)
+- What percentage of all users are active users? (How many inactive users do we have?)
```
TomasTomecek, line 22: this one will help with Franta's point from above right? when asking about installations that are not utilized

adoption-metrics/define_adoption_metrics.md
TomasTomecek: as soon as we don't have to spend a month creating that chart, I'm fine with either :D heatmap looks pretty wicked, while histogram is the classic

TomasTomecek (approved):

adoption-metrics/define_adoption_metrics.md
lachmanfrantisek: And are we going to count users or approved namespaces?

adoption-metrics/define_adoption_metrics.md
lachmanfrantisek: Yes, we just need to clearly define who is an active user, newcomer, ...

adoption-metrics/define_adoption_metrics.md
nikromen: Hmmm that's true, (for me) it makes sense to count approved namespace (I made the equation user == namespace in my head. Sorry for the confusion)

PR #146, opened 2022-06-09
research image builder

TomasTomecek:


image-builder/README.md
```diff
@@ -0,0 +1,68 @@
...
+### Usage
+
+Set the offline token via `REDHAT_API_OFFLINE_TOKEN` env var and run the
+"image_build.py" script.
+
+## Next steps
+
+The integration is pretty straightforward:
+
+1. Create a new job: "image-builder" (or "image-build"?)
```
jpopelka, line 54: For me, "image" automatically means "a container image".
When I started reading the README, I was confused about why "bootable", until I realized we're talking about VM images here.
Also, if we wanted to produce container images one day, we would need to distinguish the jobs somehow. Therefore I suggest "vm-image-build" in this case.

And +1 to `build`, not `builder`.

image-builder/image_build.py
```diff
@@ -0,0 +1,159 @@
...
+            ],
+            "customizations": {
+                "packages": ["cockpit"],
+                "payload_repositories": [
+                    {
+                        "rhsm": False,
+                        "baseurl": (
+                            "https://download.copr.fedorainfracloud.org"
+                            "/results/@cockpit/cockpit-preview/centos-stream-9-x86_64/"
+                        ),
```
lachmanfrantisek, line 104: Is it possible to be really explicit and use NVRs?

lachmanfrantisek (commented):
Looks really promissing!

image-builder/image_build.py
TomasTomecek: I'll try to do that in the packages section to pick a specific build. It should work :)

image-builder/image_build.py
```diff
@@ -0,0 +1,159 @@
...
+        while True:
+            response_json = self.image_builder_session.get(
+                f"{image_builder_api_url}/composes/{build_id}",
+            ).json()
+            status = response_json["image_status"]["status"]
+            if status not in ("building", "pending"):
+                pprint(response_json)
+                break
+            print(f"{build_id} - {status}")
+            time.sleep(3.0)
```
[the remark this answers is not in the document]
TomasTomecek: yup, polling is the only option :/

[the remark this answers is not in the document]
lachmanfrantisek: And what about the option of using celery retries for that?

image-builder/image_build.py
TomasTomecek: just confirmed this works

except that Copr only provides the latest build of package within a project

praiskup: > except that Copr only provides the latest build of package within a project

Not really, builds are removed no earlier than when they are 14 days old:
https://docs.pagure.org/copr.copr/user_documentation.html#how-long-do-you-keep-the-builds

DNF thought picks the latest-greatest NVR. If you care about particular PR-specific NVR, you
can install via fully specified NVR, like: "dnf install csdiff-2.5.0.20220615.104203.gde72e8f-1.fc36" (even thought the last version is csdiff-2.5.0.20220615.104203.gde72e8f-1.fc36.x86_64).

TomasTomecek: ah, 2 weeks, so that's why the old cockpit build could not be found, thanks for correcting me!

Image Builder uses its own layer on top of dnf so it's possible they may need to explicitly account for this use case.

image-builder/image_build.py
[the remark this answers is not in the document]
TomasTomecek: I consider that to be polling too :D although not active

My point was: there is no message bus nor webhooks/callbacks. We need to find out ourselves that an image build has finished.

PR #153, opened 2022-07-15
Workers scaling

jpopelka:


workers-scaling/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Summary/Suggestion
+
+- Default for stg & dev can stay as it is, i.e. [one worker for all tasks](https://github.com/packit/deployment/blob/cb47f9e6f806f9e340669031c7eb58df02b164f1/playbooks/deploy.yml#L40)
+  with no concurrency/autoscaling/prefetching.
+- For prod, we should experiment with setting `--autoscale` (or just `--concurrency`)
+  together with `--pool=eventlet/gevent`. We should have only [one worker for serving
+  short-running tasks](https://github.com/packit/deployment/blob/cb47f9e6f806f9e340669031c7eb58df02b164f1/vars/packit/prod_template.yml#L77)
+  which would autoscale based on load and [no worker for all tasks](https://github.com/packit/deployment/blob/cb47f9e6f806f9e340669031c7eb58df02b164f1/vars/packit/prod_template.yml#L76)
+  to avoid [p-s#375](https://github.com/packit/packit-service/issues/375) regression.
```
lachmanfrantisek, line 88: What about trying on stage first with multiple workers of the same type?

workers-scaling/README.md
```diff
@@ -0,0 +1,88 @@
...
+  the number of CPUs on host.
+- **Threads** ([eventlet](https://docs.celeryq.dev/en/latest/userguide/concurrency/eventlet.html)/[gevent](https://www.gevent.org) pool) -
+  when tasks are I/O or network bound. The number of greenlets (green threads) is
+  unrelated to the number of CPUs on host and can go even to thousands.
+
+#### Autoscaling
+
+Regarding the `--concurrency` value, we even don't have to set a static number
+if we use [worker --autoscale](https://docs.celeryq.dev/en/latest/userguide/workers.html#autoscaling)
+which dynamically resizes the pool based on load.
```
lachmanfrantisek, line 78: Does it make sense to cover also auto-scaling on openshift level?

lachmanfrantisek, line 78: Also, what about a situation where we have a huge queue of inexpensive tasks (if I get that right, this will not show up in a higher load).

workers-scaling/README.md
jpopelka: Are you suggesting first trying your packit/deployment#361 on stage?
(your packit/deployment#361 would definitely work, but it is manual and takes much more resources (x pods vs. 1 pod with x threads))

Or to try on stage some variant of what I suggest here for prod?

workers-scaling/README.md
jpopelka: Added a section.

jpopelka: The load here most likely doesn't mean CPU load, because Celery has no clue about such metric, but number of requests in the message queue - I checked the code, and Celery just [compares](https://github.com/celery/celery/blob/aa9fd8a6c06e69c7eda2a59866c3d84622c85d20/celery/worker/autoscale.py#L85) whether all threads have [something to do](https://github.com/celery/celery/blob/aa9fd8a6c06e69c7eda2a59866c3d84622c85d20/celery/worker/autoscale.py#L150) and if so and we haven't scaled to max threads, add more threads.

I updated the text.

workers-scaling/README.md
lachmanfrantisek: Sorry for late responses... I meant the second option -- try your changes on stage. (And add more workers to stage so the multi-worker processing is used.)

workers-scaling/README.md
lachmanfrantisek: Nice, thanks for the explanation!

lachmanfrantisek (approved):
Thanks!

PR #155, opened 2022-07-25
Plan triggering tests with builds from other PRs

lbarcziova:
Related to packit/packit-service#1544

### DIFF
--- a/tests-builds-from-more-prs/README.md
+++ b/tests-builds-from-more-prs/README.md
@@ -1,0 +1,89 @@
+# Triggering testing farm tests with builds from other PRs (projects)
+
+Use case: having 2 builds for different PRs, being able to trigger TF tests using both builds with a Packit
+command comment in one PR
+
+## Architecture notes:
+
+`Copr allows interlinking Copr projects so that TF could just install proper packages and Copr would enable the whole dependency chain`
+
+- it is possible to specify `runtime_dependencies` in Copr: https://pagure.io/copr/copr/pull-request/2245#
+  (should be released soon) - list of external repositories that will be automatically enabled together with this project repository
+- we would need to have a dynamic way of specifying these `runtime_dependencies`
+- since we want concrete packages to be installed in TF (we pass the build ID and NVRs), this wouldn't be enough
+  - if users build all the PRs in one Copr repo (e.g.
+    [leapp](https://github.com/oamg/leapp/blob/351cfe75ce1a7e10d7e565d9fa0bdc946a9af66d/.packit.yaml#L9) and
+    [leapp-repository](https://github.com/oamg/leapp-repository/blob/8b228941d34b3fcf4322353806fdcafa9fbba550/.packit.yaml#L26) build their PRs in one common Copr repo)
+    we cannot enforce with this solution what should be installed during tests (build ID, NVRs).
+
+## Possible implementation
+
+### Requirements
+
+The requirement would be to have enabled Packit Copr build job on both repositories where the PRs are happening.
+Then, the testing farm test job could be triggered by comment specifying the other pull request ID
+and we would rely on builds being already done by Packit.
+
+### Implementation
+
+We should have everything needed in our database and be able to get Copr build ID and package NVRs via having the other PR ID from there:
+`PullRequestModel.get_or_create(pr_id=xy, namespace=xy, repo_name=xy, project_url=xy).get_copr_builds()`
+and determine the latest build comparing one of the stored times (task_accepted, build_submitted).
+
+With having the Copr build ID and package NVRs of the other PR Copr build, we could just pass these
+into the Testing farm payload.
+
+Process:
+
+1. Try to get the PR specified in the comment from the DB - if not able to find, report this to user
+2. Get the latest Copr build for the PR from the DB - if no Copr build, report this to user
+3. Append artifact including the ID of the Copr build and NVRs to the payload for Testing Farm
+4. Trigger the tests
+5. When reporting the status, make sure it is done on the "main" PR (probably corresponding to the
+   first one build in the artifacts)
+
+#### Another option
+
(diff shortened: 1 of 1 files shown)


tests-builds-from-more-prs/README.md
```diff
@@ -0,0 +1,76 @@
...
+    [leapp-repository](https://github.com/oamg/leapp-repository/blob/8b228941d34b3fcf4322353806fdcafa9fbba550/.packit.yaml#L26) build their PRs in one common Copr repo)
+    we cannot enforce with this solution what should be installed during tests (build ID, NVRs).
+
+## Possible implementation
+
+### Requirements
+
+The requirement would be to have enabled Packit Copr build job on both repositories where the PRs are happening.
+Then, the testing farm test job could be triggered by comment specifying the other pull request ID
+and we would rely on builds being already done by Packit.
```
TomasTomecek, line 25: Sounds realistic and as a solid initial requirement

tests-builds-from-more-prs/README.md
```diff
@@ -0,0 +1,76 @@
...
+questions:
+
+1. do we want to support specifying also Git forge?
+2. do we want to do this only for PRs? (or also Copr builds from commits/releases from other repo)
+
+suggestions:
+
+1. `packit test --with-pr --pr-id 123 --repository repo --namespace namespace (--forge github.com)`
+
+2. `packit test --with-pr namespace/repo#123`
```
TomasTomecek, line 72: +1 to start here, use the native GitHub syntax and extend this in future to other gitforges (or more options)

tests-builds-from-more-prs/README.md
```diff
@@ -0,0 +1,76 @@
...
+and we would rely on builds being already done by Packit.
+
+### Implementation
+
+We should be able to get Copr build ID and package NVRs via our database having the other PR ID:
+`PullRequestModel.get_or_create(pr_id=xy, namespace=xy, repo_name=xy, project_url=xy).get_copr_builds()`
+and determine the latest build comparing one of the stored times (task_accepted, build_submitted).
+
+With having the Copr build ID and package NVRs of the other PR Copr build, we could just pass these
+into the Testing farm payload.
```
TomasTomecek, line 34: This will initially work, 100%. I'm wondering if there will be a better way to do it long-term (can Copr help us? can we just pass the Copr project to TF and let it deal with it?).

As the implementation starts, we can experiment :)

TomasTomecek (approved):

tests-builds-from-more-prs/README.md
lbarcziova: I asked Pavel about the sentence from the architecture notes and he proposed using `runtime_dependencies` which I mentioned in the doc, but with our current workflow where we specify the Copr builds in the TF payload directly via build ID and NVRs, this would not be enough. Also, I think the situation can change after resolving packit/packit-service/issues/1532.

tests-builds-from-more-prs/README.md
```diff
@@ -0,0 +1,76 @@
+# Triggering testing farm tests with builds from other PRs (projects)
+
+Use case: having 2 builds for different PRs, being able to trigger TF tests using both builds with a Packit
+command comment in one PR
```
lachmanfrantisek, line 4: Yesterday, we discussed with Tomáš, if it makes sense to cover builds as well.

Hopefully, most use-cases can be covered by a shared Copr project and (re)triggering the builds in the correct order. So it's probably not needed but we should at least have this in mind. And also, this might be a prerequisite to running the tests with correct build versions.

tests-builds-from-more-prs/README.md
```diff
@@ -0,0 +1,76 @@
...
+How can a testing farm job be retriggered:
+
+1. via comment - we would require each time to explicitly pass the PR related attributes
+2. via Github check - we would trigger the tests only with the build from the current PR
+
+### Syntax of the command
+
+questions:
+
+1. do we want to support specifying also Git forge?
```
lachmanfrantisek, line 65: I would probably don't require it but support that (similarly to the default Copr project naming schema). Depending on the difficulty, we can support both from the start or implement the other forges later on.

tests-builds-from-more-prs/README.md
```diff
@@ -0,0 +1,76 @@
...
+2. do we want to do this only for PRs? (or also Copr builds from commits/releases from other repo)
```
lachmanfrantisek, line 66: Currently, we don't support comment commands for other triggers, so:

* have it configurable
* support comment commands for commits (the question is what to do with release commits -- `release` or `commit`). 
* issues

I think this is not needed for the prototype.

tests-builds-from-more-prs/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+suggestions:
+
+1. `packit test --with-pr --pr-id 123 --repository repo --namespace namespace (--forge github.com)`
```
lachmanfrantisek, line 70: Are we speaking about CLI or comment commands?

Do we want to support CLI-like options in comments? Or use it as an argument? (e.g. `/packit test other/repo#123`)

lbarcziova: these are all suggestions for comment commands, I added here more options, we can vote about the best/combination, the `--with-pr` is not strictly needed

lachmanfrantisek: Aha, the missing `/` character confused me...

lbarcziova: forgot about that, I will update it to be clear

tests-builds-from-more-prs/README.md
lbarcziova: my original thought here was a project that has Packit configured for both PRs and commits and in some PR wants to trigger tests with a build from commit/release :thinking:

> I think this is not needed for the prototype.

:+1:

Issue #159, opened 2022-09-16
Figure out a way to proactively update configs using deprecated features


csomh:
In many cases transforming deprecated config options (like `synced_files` or `metadata`) into the current format can be done programmatically, so we could come up with a way to propose these updates as PRs to the projects still using them. This could speed up their deprecation and let us remove some code sooner.

Research how this could be done.

One of the things that needs to be done is to get an up-to-date version of all the packit.yaml-s currently in use. We discussed either saving them in the DB, or writing a script to get the list of projects and branches using the service API, and downloading the configs.

lachmanfrantisek:
ping the sync

PR #163, opened 2022-10-13
Initial research for configuration versioning

lbarcziova:
Related to packit/packit#1672

### DIFF
--- a/configuration-versioning/README.md
+++ b/configuration-versioning/README.md
@@ -1,0 +1,134 @@
+# Configuration versioning
+
+Example use cases we would like to solve with this:
+
+- Example 1: Image builder job - it will probably evolve over time depending on the requirements
+- [Example 2](https://github.com/packit/packit.dev/issues/443): Changing interpretation of job overrides from replacing the values to merging
+
+## Existing support
+
+- no support from Marshmallow, [see](https://github.com/marshmallow-code/marshmallow/issues/1107)
+- in general, I have not found any existing support to do this
+- some tips from the issue above:
+  - utilising `render_module` (from [this class](https://marshmallow.readthedocs.io/en/3.0/api_reference.html#marshmallow.Schema.Meta)) - a class that defines loads and dumps, which defaults to json,
+    could be a good place to intercept the raw data on a per-schema basis
+  - migrating the raw data before deserializing to avoid maintaining a schema history
+
+## Versioning jobs aside from the whole config
+
+- this would be mostly beneficial for versioning only one job type (e.g. vm_image_build)
+- since all the job-related fields in the job config are grouped in a class together, this would not be straightforward
+- we could have specific fields that would be versioned and have own schema
+  e.g. in case of Image builder job having some field that would introduce a nested schema
+  and we could have different versions of these: e.g.
+
+```yaml
+- job: vm_image_build
+    trigger: pull_request
+    image_builder_metadata:
+      version: 1
+      image_request:
+        packages_to_install: [packit]
+        owner: packit
+        project: packit-dev
+      image_customizations:
+        image_type: aws
+        image_architecture: x86_64
+        image_account_id: "727920394381"
+```
+
+```yaml
+- job: vm_image_build
+    trigger: pull_request
+    image_builder_metadata:
+      version: 2
+      image_distribution: rhel-8
+      image_type: aws
+      image_architecture: x86_64
+      image_account_id: "727920394381"
+      packages_to_install: [packit]
+      owner: packit
+      project: packit-dev
+```
+
+and schemas:
+
+```python
+class ImageBuilderMetadataV1(ImageBuilderMetadataSchema):
+    image_distribution = fields.String(missing=None)
+    image_request = fields.Dict(missing=None)
+    image_customizations = fields.Dict(missing=None)
+```
+
+```python
+class ImageBuilderMetadataV2(ImageBuilderMetadataSchema):
+    image_distribution = fields.String(missing=None)
+    image_type: fields.String(missing=None)
+    image_architecture = fields.String(missing=None)
+    image_account_id = fields.String(missing=None)
+    packages_to_install = fields.String(missing=None)
+    owner = fields.String(missing=None)
+    project = fields.String(missing=None)
+```
+
+and the config object would have all the fields and the code could handle both
+
+- from the coding perspective this would not be a problem - the version would be detected either from whole config
+  or directly from the `image_builder_metadata`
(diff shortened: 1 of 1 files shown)


configuration-versioning/README.md
```diff
@@ -0,0 +1,137 @@
...
+- this would be properly documented, we can get inspired by Docker,
+  [see](https://docs.docker.com/compose/compose-file/compose-versioning/#versioning), the documentation
+  would also include Packit version that starts supporting the particular configuration schema
+
+### Having more schemas vs having one schema
+
+#### Having more schemas
+
+- maintaining more versions of schema for whole package config should not be needed
+  - the schemas would be mostly have the same content, only with difference in few fields
```
jpopelka, line 96:
```suggestion
  - the schemas would mostly have the same content, only with differences in a few fields
```

jpopelka (commented):
Have you perhaps seen any project using this versioning of specific parts of config?

configuration-versioning/README.md
```diff
@@ -0,0 +1,137 @@
...
+
+- we would need to make sure that load-dump-load results in same representation
+
+## Plan
+
+- decide whether/to which scope we want to do this in relation to [this issue](https://github.com/packit/research/issues/159)
+  - if we would introduce some change that would break the compatibility, we would open PRs for each repository that
+    uses the particular configuration field
+  - this could not be sufficient for all use cases
+- decide whether we want to version whole config or just have versioned fields or use combination
```
TomasTomecek, line 137: Versioning jobs or sections sound appealing although I am concerned about usability for users. Even right now, many of our users still use deprecated options (including us!) so we should minimize the number of changes needed to be done to packit.yaml.

Once Hunor finishes his config changes related to monorepo, I wonder how will those changes impact this research.

configuration-versioning/README.md
```diff
@@ -0,0 +1,137 @@
...
+                ...
+
+        return data
+```
+
+- we would need to make sure that load-dump-load results in same representation
+
+## Plan
+
+- decide whether/to which scope we want to do this in relation to [this issue](https://github.com/packit/research/issues/159)
```
FrNecas, line 133: I feel like that as a part of #159 we will be able to do mainly simple changes (such as renaming). There may, however, be more complex changes to be made which we can't do automatically, as you mention later. I see one slight conflict of this with versioning. Imagine the following scenario:

1) user has a config of version `1`
2) we make a backwards incompatible change which cannot be handled automatically for users using this feature, they need to update manually to schema version `2`. But let's suppose that the user doesn't do this
3) We rename a field in the config, resulting in schema version `3`. We would be able to create a PR for this but we can't. Going version `1` -> `3` could break things because we can't "migrate" to changes done in `2`.

FrNecas: That shouldn't really be a problem to implement but it will be another requirement for the automatic updates to take this into consideration

lbarcziova: yes, you are right, I just wanted to mention this in the research since it is somehow related

lbarcziova:
> Have you perhaps seen any project using this versioning of specific parts of config?

No, I haven't, I only considered this options since it was suggested in the issue

configuration-versioning/README.md
lbarcziova: > Versioning jobs or sections sound appealing although I am concerned about usability for users.

I agree, it could be simpler from an implementation pov, but very confusing to users.

lbarcziova:
Should I research something else here?

configuration-versioning/README.md
csomh: Versioning sections of the schema has the strong potential of creating a very confusing and difficult to follow environment both for developers and users. I'm strongly against going in this direction.

configuration-versioning/README.md
```diff
@@ -0,0 +1,137 @@
...
+- in general, I have not found any existing support to do this
+- some tips from the issue above:
+  - utilising `render_module` (from [this class](https://marshmallow.readthedocs.io/en/3.0/api_reference.html#marshmallow.Schema.Meta)) - a class that defines loads and dumps, which defaults to json,
+    could be a good place to intercept the raw data on a per-schema basis
+  - migrating the raw data before deserializing to avoid maintaining a schema history
+
+## Versioning jobs aside from the whole config
+
+- since we have the inheritance CommonPackageConfig <- PackageConfig and JobConfig and all the job-related fields
+  in the job config together, this would be really messy
```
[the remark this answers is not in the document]
csomh: Yes, this was changed. There is no inheritance between these classes.

configuration-versioning/README.md
csomh: Another option is to drop #159 once versioning is implemented.

configuration-versioning/README.md
```diff
@@ -0,0 +1,141 @@
...
+        add all the fields (except for jobs) to every job so we can process only jobconfig in p-s
+        """
+        if not data:  # data is None when .packit.yaml is empty
+            return data
+        for job in data.get("jobs", []):
+            for k, v in data.items():
+                if k == "jobs":
+                    # overriding jobs doesn't make any sense
+                    continue
+                if self.version == 1:
```
csomh, line 123: Doing this in the `pre_load` step seems late to me, and will make code more and more complicated over time.

I would rather do something like: check the version in `PackageConfig.get_from_dict()` (default to some pre-defined version if the version is not specified – the last one could be a good candidate) and then import the schema classes matching that version:

```python
if version == "1":
    from packit.schema.v1 import PackageConfigSchema
elif version == "2":
    from packit.schema.v2 import PackageConfigSchema
```

This would allow partially redefining schema classes or reexporting existing schema classes. For example in `packit.schema.v2` one could have:

```python
from packit.schema.v1 import PackageConfigSchema as PackageConfigSchemaV1
# reexport as v2, without a change
from packit.schema.v1 import JobConfig

class PackageConfigSchema(PackageConfigSchemaV1):
    # override some fields
```

Would something like this make sense?

csomh:
Will there be also a need to version the config classes? Or versioning the schema would be enough?

configuration-versioning/README.md
lbarcziova: yes, this makes sense, I will incorporate it into the document

lbarcziova:
> Will there be also a need to version the config classes? Or versioning the schema would be enough?

My way of thinking was to only version schema and then always have only one config class for the particular config (` all versions of the configuration need to be migrated when loading the configuration to the latest state of the configuration object`, I will rephrase this to make it more clear)

PR #170, opened 2023-01-27
Monorepo code refactoring research

majamassarini:


csomh (approved):
Thanks for doing the research and sharing your findings.

I agree with the conclusion: the decorator approach will probably be the lowest impact in terms of the amount of code to change, and it feels that it will be possible to reuse it in Packit Service as well.

majamassarini:
> Thanks for doing the research and sharing your findings.

Thanks for your support and reviews!

lachmanfrantisek (approved):
Thank you for the thorough investigation and for trying various ways how to do that!

monorepo-support/refactoring/README.md
```diff
@@ -0,0 +1,66 @@
...
+2. For a `JobConfig` referencing multiple packages create as many `JobConfigView`, [here a draft example](https://github.com/majamassarini/packit/blob/multiple_distgit_external_package_config/packit/config/job_config.py#L108-L143), as the packages and **group them together**
+
+   Substitute the `self.event.package_config.jobs` with a call like the `package_config.get_grouped_job_views` [here](https://github.com/majamassarini/packit/blob/multiple_distgit_external_package_config/packit/config/package_config.py#L174-L196).
+   Modify `steve_job.process_jobs`, `steve_job.get_handlers_for_event`, `steve_job.get_config_for_handler_kls` methods to work with the new data structure returned by the `package_config.get_grouped_job_views`.
+
+   At the end the `steve_job.process_jobs` will create only handlers taking a list of `JobConfig` or `JobConfigView` and for this reason we will modify all our handlers to loop over all the given configs.
+
+   - **PROS**: one single way to _match jobs and handlers_
+
+   - **CONS**: we are suggesting that all the handlers should be able to handle multiple configs, but this is probably not true.
```
csomh, line 64: By this you mean that "handlers should be able to work with (build) multiple packages from a monorepo", right?

monorepo-support/refactoring/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+Personally I prefer solution 1. I see it as more simple and explicit. We probably could/should re-think the job/handler matching code from the point of view of a monorepo. I feel like it is worth doing that when we have a clear understanding of what we need in a handler capable of supporting multiple packages.
```
csomh, line 66: Hey @praiskup, I have a question for you :slightly_smiling_face:: when building monorepos in Copr, how are packages built? Is every package from the monorepo built as part of the same Copr build, or separate ones? Is there even such as thing as building monorepos in Copr?

monorepo-support/refactoring/README.md
majamassarini: Yes, right

csomh: I'm wondering if it wouldn't be beneficial to update handlers so that they are able to handle one or more configs (and so one or more packages from a monorepo). Let's see how the discussion below with Pavel turns out. If there is a use-case to build multiple packages in the same Copr build (and then test them together), then maybe we should make this change.

PR #171, opened 2023-02-07
Research GitHub API rate limits

csomh:
Calculate the approximate number of calls currently made in order to be able to asses the current situation and tell the impact of the proposed approaches. [Direct link to the rendered Jupyter Notebook](https://github.com/csomh/research/blob/github-api-call-inventory/github-api-call-inventory/github-api-call-inventory.ipynb).

Propose approaches to collect metrics.

Related to packit/packit-service#1765

### DIFF
--- a/github-api-call-inventory/README.md
+++ b/github-api-call-inventory/README.md
@@ -1,0 +1,60 @@
+# GitHub API Call Rate Limits
+
+## Inventory of GitHub API calls made during a pipeline run
+
+See [this Jupyter Notebook](./github-api-call-inventory.ipynb).
+
+## Collecting metrics on GitHub API calls
+
+We want to count the number of GitHub API calls, so that we can produce an
+indicator about the average API calls per pipeline run. Such an indicator
+could help us notice if code changes result in an increase in the API calls
+made, and so making it more likely that the call rate limits are reached in
+certain orgs.
+
+Packit and Packit-Service interact with GitHub through ogr, which in turn
+relies on PyGithub.
+
+### Approach 1: Patch PyGithub to inject a metric object
+
+The first thing, that was explored is if PyGithub could be patched in some way
+in order to enable Prometheus metrics collection. This would be possible, by
+patching [Requester] to accept a Prometheus metric object and use it. This
+then could be used to pass down a metrics collector object (for example a
+Counter) from PackitAPI through ogr to PyGithub.
+
+Going through the whole flow:
+
+In packit-service GithubProjects are obtained with
+`ServiceConfig.get_project()`. This method (implemented in
+`packit.config.config.Config`) will need to be updated to take a
+`github_call_counter` object and pass it down to `ogr.get_project()`.
+
+Here `ogr.factory.get_project()` would pass it down to `GithubService` which
+then would pass it down to `GithubProject`, which then would pass it down to
+the PyGithub `Github` object, so that it can be used as [a new property of the
+`Requester`], in order to be able to pass it to the connection, and
+incremented every time [a response is received].
+
+Patching PyGithub could be done while building the service and the worker
+images. The patches could be also submitted upstream, to get some feedback on
+the approach and maybe incorporate them in PyGithub.
+
+### Approach 2: Use a proxy
+
+An another solution that _might_ work is to update the hosts files in the
+service and worker images and route all connections to GitHub through an
+[NGINX proxy pod], and then use [NGINX Prometheus Exporter] to expose some
+[HTTP metrics], which then could be interpreted as metrics on GitHub API
+calls. Exposing these metrics—so that they can be scraped—would require
+setting up a new route, something like `github-calls.packit.dev`. I did not
+try this approach, so some more time should be spent on this, to figure out if
+it would indeed work.
+
+[requester]: https://github.com/PyGithub/PyGithub/blob/master/github/Requester.py
+[nginx proxy pod]: https://docs.nginx.com/nginx/admin-guide/web-server/reverse-proxy/
+[nginx prometheus exporter]: https://github.com/nginxinc/nginx-prometheus-exporter
+[http metrics]: https://github.com/nginxinc/nginx-prometheus-exporter#http
+[a new property of the `requester`]: https://github.com/PyGithub/PyGithub/blob/e414c3227bb15819b443b4474f1aded433011bda/github/MainClass.py#L122
+[connection]: https://github.com/PyGithub/PyGithub/blob/e414c3227bb15819b443b4474f1aded433011bda/github/Requester.py#L668
+[a response is received]: https://github.com/PyGithub/PyGithub/blob/e414c3227bb15819b443b4474f1aded433011bda/github/Requester.py#L129
--- a/github-api-call-inventory/github-api-call-inventory.ipynb
+++ b/github-api-call-inventory/github-api-call-inventory.ipynb
@@ -1,0 +1,720 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "59428441",
+   "metadata": {},
+   "source": [
+    "## Tasks"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "id": "8c84b475",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from dataclasses import dataclass, fields\n",
+    "\n",
+    "@dataclass\n",
+    "class Task:\n",
+    "    is_project_private: int\n",
+    "    is_user_approved: int\n",
+    "    load_packit_yaml: int\n",
+    "    set_status: int\n",
+    "        \n",
+    "    @property\n",
+    "    def total(self):\n",
+    "        return sum(getattr(self, field.name) for field in fields(self.__class__))\n",
+    "    \n",
+    "    def __mul__(self, num):\n",
+    "        if not isinstance(num, int):\n",
+    "            raise TypeError(f\"unsupported operand type(s) for *: {type(self)!r} and {type(num)!r}\")\n",
+    "        return Task(*[num * getattr(self, field.name) for field in fields(self.__class__)])\n",
+    "    \n",
+    "    __rmul__ = __mul__\n",
+    "    \n",
+    "    def __add__(self, obj):\n",
+    "        if not isinstance(obj, Task) and not isinstance(obj, int):\n",
+    "            raise TypeError(f\"unsupported operand type(s) for +: {type(self)!r} and {type(obj)!r}\")\n",
+    "        if isinstance(obj, int) and obj != 0:\n",
+    "            raise TypeError(f\"adding {obj!r} to an {type(self)!r} is not supported\")\n",
+    "        if isinstance(obj, int) and obj == 0:\n",
+    "            obj = Task(*[0] * len(fields(self.__class__)))\n",
+    "        return Task(*[(getattr(self, field.name) + getattr(obj, field.name)) for field in fields(self.__class__)])\n",
+    "    \n",
+    "    __radd__ = __add__"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "464ee1d0",
+   "metadata": {},
+   "source": [
+    "## Estimating the number of API calls"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "id": "bd8ae47b",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from collections import namedtuple\n",
+    "\n",
+    "metrics = [\n",
+    "    \"prs_per_hour_per_org\", \n",
+    "    \"total_is_project_private\", \n",
+    "    \"total_is_user_approved\", \n",
+    "    \"total_load_packit_yaml\", \n",
+    "    \"total_set_status\",\n",
+    "    \"grand_total\",\n",
+    "]\n",
+    "\n",
+    "class APICallMetrics(namedtuple(\"APICallMetrics\", metrics)):\n",
+    "    def __repr__(self):\n",
+    "        return \"\\n\".join(f\"{field} = {getattr(self, field)}\" for field in self._fields)\n",
+    "    \n",
+    "def github_api_calls(\n",
+    "    # The number of chroots configured to be built in Copr\n",
+    "    chroots=3, \n",
+    "    # GitHub server-to-server rate limit per hour\n",
+    "    rate_limit=5000,\n",
+    "    # Calls to load .packit.yaml.\n",
+    "    # The current implementation cycles through a list of 7 files.\n",
+    "    # If the config file is .packit.yaml, this is loaded from 2 calls.\n",
+    "    # If the config file is packit.yaml, it's loaded from 5 calls.\n",
+    "    # The +1 call is to try to find the specfile path.\n",
+    "    load_packit_yaml=2+1,\n",
+    "    # Whether builds are enabled\n",
+    "    builds_enabled=True, \n",
+    "    # Whether tests are enabled\n",
+    "    tests_enabled=True, \n",
+    "    # Load packit.yaml once during the pipeline run\n",
+    "    load_packit_yaml_once=False, \n",
+    "    # Set all flags (including test ones) when the build status changes\n",
+    "    all_flags=True,\n",
+    "    # Check if the repo is private only in the beginning\n",
+    "    is_private_once=False,\n",
+    "):\n",
+    "    build_status_flag = int(builds_enabled)\n",
+    "    test_status_flag = int(tests_enabled)\n",
+    "    tasks = {\n",
+    "        \"PR is updated\": Task(\n",
+    "            # https://github.com/packit/packit-service/blob/bc1f66bb82aa7c013bde8b8e670ce7e02f55de41/packit_service/worker/jobs.py#L156\n",
+    "            is_project_private=1,\n",
+    "            # https://github.com/packit/packit-service/blob/bc1f66bb82aa7c013bde8b8e670ce7e02f55de41/packit_service/worker/allowlist.py#L275\n",
+    "            is_user_approved=2,\n",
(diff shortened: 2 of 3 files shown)


github-api-call-inventory/github-api-call-inventory.ipynb
```diff
@@ -0,0 +1,520 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from dataclasses import dataclass, fields\n",
+    "\n",
+    "@dataclass\n",
+    "class Event:\n",
+    "    is_project_private: int\n",
+    "    load_packit_yaml: int\n",
+    "    set_status: int\n",
```
csomh, line 24: Is the GitHub API called for any other reason while handling PR events?

github-api-call-inventory/github-api-call-inventory.ipynb
```diff
@@ -0,0 +1,520 @@
...
+    "        \"SRPM build finished\": chroots * event,\n",
+    "        \"RPM build started\": chroots * event,\n",
+    "        \"RPM build finished\": chroots * event,\n",
+    "    })\n",
+    "    if tests_enabled:\n",
+    "        event = Event(is_private, load_packit_yaml, test_status_flag)\n",
+    "        events.update({\n",
+    "        \"Tests started\": chroots * event,\n",
+    "        \"Tests finished\": chroots * event,\n",
+    "    })\n",
```
csomh, line 106: Are there any other events received during a pipeline run?

lbarcziova: I think "SRPM build in Copr was submitted..." is the only missing

github-api-call-inventory/github-api-call-inventory.ipynb
lbarcziova: iirc we also check whether the one who triggered the event has write access to repo or is author of the PR

github-api-call-inventory/github-api-call-inventory.ipynb
csomh: Is that a separate event, or that's the status which is set when the update event for the PR is received?

jpopelka (commented):
:+1:

github-api-call-inventory/github-api-call-inventory.ipynb
csomh: Where is this happening? I've looked in the packit-service code, but I was able to find such check only for vm_image jobs. What am I missing? :thinking:

github-api-call-inventory/github-api-call-inventory.ipynb
lbarcziova: it is not a separate event, but it happens in a new task:
we receive a Github PR-related event, we create the "Task was accepted status"  ( + do the config loading, private check,..) and then create a new specific task and there we create the "SRPM build in Copr was submitted..." status

github-api-call-inventory/github-api-call-inventory.ipynb
csomh: Found [it](https://github.com/packit/packit-service/blob/b946857a437b062d6e93021b5420bf46f3ce9964/packit_service/worker/jobs.py#L385).

csomh: Added.

github-api-call-inventory/github-api-call-inventory.ipynb
csomh: Thanks, updated this.

github-api-call-inventory/github-api-call-inventory.ipynb
```diff
@@ -0,0 +1,765 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "0c05d101",
+   "metadata": {},
+   "source": [
+    "The main issue seems to be that `task.steve_jobs.process_message` re-creates the context of the pipeline by getting information from GitHub: checks if the project is private, if the user is approved to run the pipeline, and load the package config (`.packit.yaml`).\n",
+    "\n",
+    "All this could be avoided by looking up the pipeline when an event arrives (Copr events, testing farm events - when these are received there already needs to be a pipeline created, otherwise they are invalid), and getting this information from the pipeline.\n",
+    "\n",
+    "The only time it makes sense to load this from GitHub is when the pipeline is started, for example when the PR event is received."
```
csomh, line 458: Is this observation correct? Would be nice if someone else could confirm this, b/c otherwise the numbers are way too wrong :sweat_smile:

TomasTomecek: they shouldn't change within a pipeline run, right? people can become collaborators, be approved, packit.yaml can change, but NOT within a pipeline run: +1 to cache all of that data

this will also be a massive speedup

github-api-call-inventory/github-api-call-inventory.ipynb
```diff
@@ -0,0 +1,765 @@
...
+    "    builds_enabled=True, \n",
+    "    # Whether tests are enabled\n",
+    "    tests_enabled=True, \n",
+    "    # Load packit.yaml once during the pipeline run\n",
+    "    load_packit_yaml_once=False, \n",
+    "    # Set all flags (including test ones) when the build status changes\n",
+    "    all_flags=True,\n",
+    "    # Check if the repo is private only in the beginning\n",
+    "    is_private_once=False,\n",
+    "    # Check if a user is allowed to trigger the pipeline only once\n",
```
lbarcziova, line 101: maybe I am mistaken, but I think the user check is done only once

github-api-call-inventory/github-api-call-inventory.ipynb
```diff
@@ -0,0 +1,765 @@
...
+    "    user_check_once=True,\n",
+    "    is_private_once=True)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "7fed3b7e",
+   "metadata": {},
+   "source": [
+    "From this point on, the number of API calls could be further reduced only by having fewer status flags (merging flags for different chroots) or making fewer updates to these flags. Both of these could make the user experiance worse."
```
lbarcziova, line 724: agreed, I would not try to reduce these for the cost of worse UX

github-api-call-inventory/github-api-call-inventory.ipynb
```diff
@@ -0,0 +1,765 @@
...
+    "    user_check_once=True,\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "2284d964",
+   "metadata": {},
+   "source": [
+    "**Don't update the test flags while the build is in progress**, to get to 47 PRs/hour/org (a 327% increase in total), thanks to shaving of 24 API calls to update the status flags."
```
lbarcziova, line 550: I'm not sure about the UX here, but this could somehow help with the situation with GitLab statuses where we cannot overwrite "running" -> "running" status (packit/packit-service#1821): this way, we could have only "In progress" status pointing to the build/test correctly

lbarcziova (approved):
very nice write-up, thanks for all the numbers, this can definitely help with prioritising the improvements!

I think it would make sense to start with optimisations for packit.yaml loading and private check

github-api-call-inventory/github-api-call-inventory.ipynb
csomh: Right, Copr, Koji and TestingFarm event's [don't trigger this check](https://github.com/packit/packit-service/blob/bc1f66bb82aa7c013bde8b8e670ce7e02f55de41/packit_service/worker/allowlist.py#L237). Thanks for pointing it out!

github-api-call-inventory/github-api-call-inventory.ipynb
csomh: If GitLab has a "pending" status, and supports updates to that, then yes, this could help. Then the "running" status would be set only when the test is actually triggered.

lbarcziova: yes, GitLab has a "pending" status

csomh:
I wrote down some ideas on collecting metrics on GitHub API Calls. This is ready for final review and approval now.

github-api-call-inventory/README.md
```diff
@@ -0,0 +1,37 @@
...
+PyGithub.
+
+An another solution that _might_ work is to update the hosts files in the
+service and worker images and route all connections to GitHub through an
+[NGINX proxy pod], and then use [NGINX Prometheus Exporter] to expose some
+[HTTP metrics], which then could be interpreted as metrics on GitHub API
+calls. Exposing these metrics—so that they can be scraped—would require
+setting up a new route, something like `github-calls.packit.dev`. I did not
+try this approach, so some more time should be spent on this, to figure out if
+it would indeed work.
```
csomh, line 32: Would be nice to get a second pair of eyes on this if it indeed makes sense :smiley:

github-api-call-inventory/github-api-call-inventory.ipynb
```diff
@@ -0,0 +1,720 @@
...
+   "source": [
+    "From this point on, the number of API calls could be further reduced only by having fewer status flags (merging flags for different chroots) or making fewer updates to these flags. Both of these could make the user experiance worse."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "8745cba8",
+   "metadata": {},
+   "source": [
+    "Note, that the above assumes that pipelines are never running in parallel, which is probably not the case, for example, when PRs are updated while a previous pipeline is still in progress. In these situations old pipeline runs could still consume from the GitHub API rate limit (and maybe override statuses?). This could be solved by canceling Copr builds and test runs belonging to pipelines which become irrelevant ([packit/packit-service#5](https://github.com/packit/packit-service/issues/5))."
```
lachmanfrantisek, line 687: Agree, this can be a big help, but a) not easy to do b) not always wanted (e.g. with branch commit push, I don't think it's clear if the user wants to skip the build triggered by a previous commit in case of it still being in progress).

Do you think it's worth investigating? Or, it's more beneficial to spend our time on other suggestions you've found?

github-api-call-inventory/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+Packit and Packit-Service interact with GitHub through ogr, which in turn
+relies on PyGithub.
+
+The first thing, that was explored is if PyGithub could be patched in some way
+in order to enable Prometheus metrics collection. This would be actually
+possible, by patching [Requester] to make it possible to accept a Prometheus
+metric object and use it, but I couldn't figure out an easy way to create this
+object in Packit and Packit-Service, pass it to ogr, and then pass it to
+PyGithub.
```
lachmanfrantisek, line 23: @mfocko Do you think we will be able to implement such feature into OGR?

If we don't need to be precise, I think we can try collecting function calls of pygithub -- no 1-1 mapping to remote calls, but at least a rough idea of where to look closer.

github-api-call-inventory/README.md
lachmanfrantisek: Will this "man-in-the-middle" work?

If yes, then it looks like the cleanest solution without any unreliable library patching...

github-api-call-inventory/README.md
lachmanfrantisek: It reminds me of our beginnings with ogr and requre when we were patching off all the calls... Actually, we can try using requre for this since there is a way how to catch all the remote calls on `requests` level....;)

lachmanfrantisek (approved):
Wow! Thanks for such a detailed investigation! I'm really glad you haven't done any hacky solution for the original issue instead of this! Thanks!

github-api-call-inventory/github-api-call-inventory.ipynb
```diff
@@ -0,0 +1,720 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "616bb63c",
+   "metadata": {},
+   "source": [
+    "Let's pick the worst case from the list above (*6 chroots, config is called 'packit.yaml'*) as a reference, and check how the number of API calls change if we apply the above.\n",
+    "\n",
+    "**Load the package config only once per pipeline run** to increase the number of possible runs from 14 to 38 (~171% increase). This is achived by reducing the number of API calls to get the package config from 222 to 6."
```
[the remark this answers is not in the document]
csomh: I'm not suggesting anything :hand_over_mouth: But yes, probably that'll be the solution.

github-api-call-inventory/github-api-call-inventory.ipynb
```diff
@@ -0,0 +1,720 @@
...
+    "    load_packit_yaml_once=True)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "2284d964",
+   "metadata": {},
+   "source": [
+    "**Don't update the test flags while the build is in progress**, to get to 47 PRs/hour/org (a 235% increase in total), thanks to shaving of 24 API calls to update the status flags."
+   ]
```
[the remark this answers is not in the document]
csomh: Currently when build events are handled both build flags (ex: `rpm-build:fedora-37-x86_64`) and test flags (ex: `testing-farm:fedora-37-x86_64`) are updated.

This means that test flags will cycle through the same 

1. "SRPM build in Copr was submitted"
1. "SRPM build in Copr has started..."
1. "SRPM build in Copr has finished."
2. "RPM build is in progress..."
3. "RPMs were built successfully."

messages. Setting each of these is another API call. Instead of doing this when the SRPM build in Copr is submitted, test flags could be set to something like "Waiting for RPMs to be built...", and updated only when tests are submitted.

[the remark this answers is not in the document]
lachmanfrantisek: Yes, that's what I meant. Thanks and agree.

github-api-call-inventory/github-api-call-inventory.ipynb
csomh: Why would it be difficult?

We have pipelines and from those we should know which builds/tests are still running. We could cancel those, by implementing something like `pipeline.cancel()`. 

Canceling previous pipelines could be optional, or not done at all in the case of branch pushes. In the case of PRs though it doesn't make sense to let old pipelines to finish.

github-api-call-inventory/README.md
csomh: I'll try to come up with a small local experiment to try this out.

github-api-call-inventory/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+See [this Jupyter Notebook](./github-api-call-inventory.ipynb).
+
+## Collecting metrics on GitHub API calls
+
+We want to count the number of GitHub API calls, so that we can produce an
+indicator about the average API calls per pipeline run. Such an indicator
+could help us notice if code changes result in an increase in the API calls
+made, and so making it more likely that the call rate limits are reached in
+certain orgs.
```
TomasTomecek, line 13: I'd also appreciate seeing the status codes of those requests so we know how many of them are `ok`, `4xx`, `2xx`

github-api-call-inventory/README.md
TomasTomecek: whaaaat, I would love to see this experiment 😍

if it proves reliable, we could use it for other services as well

TomasTomecek (approved):

github-api-call-inventory/README.md
csomh: It seems the proxy approach would be more complicated than I assumed. NGINX doesn't support forward proxying out of the box (in this case, this is what is needed, not the reverse proxy feature). Solutions do exist, but they involve patching NGINX code, which I consider quite hacky.

I've started looking at PyGithub again, and searching for a way to inject some response hooks into the `requests` session in `github.Requester`, and figure out a way to reliably pass down these hooks from ogr objects. Metrics collections could happen in these hook functions, and we could also make @TomasTomecek happy and create metrics for (groups of) status codes :slightly_smiling_face:

csomh:
I've updated this to describe a rough approach for how to patch PyGithub in order to be able to use a metric collection object for counting calls. Please take another look at "Approach 1".

TomasTomecek (approved):
Wow, the numbers are just mind-blowing.

github-api-call-inventory/github-api-call-inventory.ipynb
lachmanfrantisek: (Sorry for not getting back to this earlier..)

> Why would it be difficult?

I am not saying difficult, just that it requires a non-trivial change. (Compared to other enhancements that are more isolated.)

lachmanfrantisek: So, not against it, but not sure if we should spend our time implementing that...;)

github-api-call-inventory/README.md
```diff
@@ -0,0 +1,56 @@
...
+In packit-service GithubProjects are obtained with
+`ServiceConfig.get_project()`. This method (implemented in
+`packit.config.config.Config`) will need to be updated to take a
+`github_call_counter` object and pass it down to `ogr.get_project()`.
+
+Here `ogr.factory.get_project()` would pass it down to `GithubService` which
+then would pass it down to `GithubProject`, which then would pass it down to
+the PyGithub `Github` object, so that it can be used as [a new property of the
+`Requester`], in order to be able to pass it to the connection, and
+incremented every time [a response is received].
```
lachmanfrantisek, line 37: Sounds good!

lachmanfrantisek: What do you suggest from the deployment point of view?

* A PyGithub source-git?
* Do the patching inside image-build?

Maybe, we can try to submit this upstream since this might be relevant for others as well... But I would expect if this got refused...

lachmanfrantisek (approved):
Thanks again Hunor!

github-api-call-inventory/README.md
csomh: I would apply the patches inside the image build, to keep things simple. And yes, they could be submitted upstream, to get community feedback on it.

csomh: Made a note about this.

lachmanfrantisek: Thanks!

PR #172, opened 2023-02-14
How to make packit-service-tests-openshift OpenShift-less?

jpopelka:


csomh:
By using `oc cluster up` the deployment playbooks were also tested, which was useful, b/c we don't run them that often—running them in the tests made them being used on a regular basis and issues could be surfaced early on. Will this change when using `kube play`?

jpopelka:
Good point, no, the deployment playbooks would not be tested.

In an ideal world, they would be tested in the deployment repo, but there's still the packit/deployment#310 which prevents us from running the tests there - I'm giving it another try - packit/deployment#455

jpopelka:
> In an ideal world, they would be tested in the deployment repo, but there's still the [packit/deployment#310](https://github.com/packit/deployment/issues/310) which prevents us from running the tests there - I'm giving it another try - [packit/deployment#455](https://github.com/packit/deployment/pull/455)

Fixed by packit/deployment#455, packit/packit-service#1925, packit/tokman#55, packit/packit-service-fedmsg#79

PR #185, opened 2023-05-17
Document how to count the number of patches in packages

csomh:
This is to document the research done to better understand how patches are used across different distributions in the RHEL-family, and to save the scripts written  during this research for future use.

### DIFF
--- a/packages-with-patches/README.md
+++ b/packages-with-patches/README.md
@@ -1,0 +1,80 @@
+# Counting patches in specfiles
+
+Ways to count the number of patches in the packages in a distribution.
+
+## Fedora Rawhide
+
+Download and unpack the specfile archive for the packages currently in
+Rawhide:
+
+```
+$ curl -LO https://src.fedoraproject.org/repo/rpm-specs-latest.tar.xz
+$ tar -xvf rpm-specs-latest.tar.xz
+```
+
+Count the number of of specfiles with a given number of patches:
+
+```
+$ grep -crE '^Patch[0-9]*:' rpm-specs/ | grep --only-matching -E '[0-9]+$' | sort -g | uniq -c
+```
+
+The command above tells how many specfiles there are with 0, 1, 2, ...
+patches.
+
+## CentOS Stream
+
+Use `download-stream.py` and specify the stream from which specfiles should be
+downloaded. This uses GitLab.com API, so you'll need to specify the token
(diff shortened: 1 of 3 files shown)


packages-with-patches/download-rhel.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+def any_branch(url, branches):
+    """Check if a repo on a given URL has at least one of the branches
+
+    Args:
+        url: URL of the repo
+        branches: list of branches to be checked
+    Returns:
+        The name of the branch which is found the first or None if none of
```
lbarcziova, line 46: doesn't it return the last found branch?

lbarcziova (approved):
thanks for writing this down!

packages-with-patches/download-rhel.py
csomh: Right, I fixed the doc-string and also how to call this script, when there are multiple branches to pick from.

(Originally I started with going from the latest version to the newest, but then I realised it would result in fewer calls—and shorter execution time—if it's done the other way around.)

packages-with-patches/download-rhel.py
```diff
@@ -0,0 +1,129 @@
...
+def search_spec(url, branch):
+    tree_url = f"{url}tree/?h={branch}"
+    response = requests.get(tree_url)
+    response.raise_for_status()
+    content = response.content.decode()
+    for line in content.splitlines():
+        match = SPECFILE.match(line)
+        if match:
+            spec_url = match.group("spec_url").replace("tree/", "plain/")
+            response = requests.get(urljoin(CGIT, spec_url))
```
TomasTomecek, line 65: I like this is not cloning the whole repos and only fetches the spec.

TomasTomecek (approved):
nicely done

packages-with-patches/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+```
+$ curl -LO https://src.fedoraproject.org/repo/rpm-specs-latest.tar.xz
+$ tar -xvf rpm-specs-latest.tar.xz
+```
+
+Count the number of of specfiles with a given number of patches:
+
+```
+$ rg -c --include-zero '^Patch\d*:' rpm-specs/ | rg --only-matching '\d+$' | sort -g | uniq -c
```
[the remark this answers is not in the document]
csomh: Updated to use `grep` instead.

PR #187, opened 2023-05-29
Research DB composite types

lbarcziova:
Fixes https://github.com/packit/packit-service/issues/1985

### DIFF
--- a/database-composite-types/README.md
+++ b/database-composite-types/README.md
@@ -1,0 +1,174 @@
+# Usage of composite type to store the pipelines
+
+[Composite column types](https://docs.sqlalchemy.org/en/14/orm/composites.html):
+Sets of columns can be associated with a single user-defined datatype, which in modern use is normally a Python dataclass. The ORM provides a single attribute which represents the group of columns using the class you provide.
+
+Examples how can this be done:
+
+1. sqlalchemy_utils [CompositeType](https://sqlalchemy-utils.readthedocs.io/en/latest/data_types.html?highlight=CompositeType#module-sqlalchemy_utils.types.pg_composite):
+   a custom SQLAlchemy type designed to work with PostgreSQL's composite types:
+
+```python
+class MyCompositeType(CompositeType):
+    attribute1 = Column(Integer)
+    attribute2 = Column(String)
+
+class MyModel(Base):
+    __tablename__ = 'my_table'
+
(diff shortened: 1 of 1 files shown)


database-composite-types/README.md
```diff
@@ -0,0 +1,172 @@
+# Usage of composite type to store the pipelines
+
+[Composite column types](https://docs.sqlalchemy.org/en/20/orm/composites.html):
```
jpopelka, line 3: The link is to SQLAlchemy 2.0, while we're currently using 1.4 (and if we move our images from Fedora to CentOS Stream 9 we'll stay with 1.4 for quite some time), so I'd use the 1.4 version of the docs. 

```suggestion
[Composite column types](https://docs.sqlalchemy.org/en/14/orm/composites.html):
```

Does that affect your examples?

database-composite-types/README.md
```diff
@@ -0,0 +1,172 @@
...
+```
+
+## 2.option
+
+How could the model look like when using composite type on group level:
+
+- "groups" stored together in one pipeline
+- mapping build -> TF needed
+
+### Issues
```
jpopelka, line 160: I see only issues, are there any pros as well? :wink:

lbarcziova: sorry, I thought I included it in the beginning (added now); as the issue suggests, the idea was to make the database design pipeline-centric and remove the need to do the joins on multiple tables

database-composite-types/README.md
jpopelka: You marked this as resolved without touching the link or commenting, so unresolving :stuck_out_tongue_winking_eye:

database-composite-types/README.md
```diff
@@ -0,0 +1,174 @@
...
+Examples of when this happens:
+
+- when updating Copr build status in DB, we get the corresponding builds via the build ID from Copr (index)
+- when updating TF run status in DB, we get the corresponding test run via the TF pipeline ID (index)
+- babysit tasks - they get all the pending Copr builds / Test runs
+- when triggering `/packit test` - we get the latest Copr build model with corresponding commit SHA
+
+#### 2. data duplication
+
+- updating SRPM build would require getting all pipelines and updating the data everywhere
```
majamassarini, line 122: I agree that data duplication is not a good thing.

majamassarini (approved):

PR #191, opened 2023-07-17
feat: adjust for inclusion on packit.dev

mfocko:


jpopelka (approved):

README.md
```diff
@@ -1,44 +0,0 @@
-# Research notes
```
TomasTomecek, line 1: some documentation would be highly appreciated to understand how to work with this repo since readme is being dropped

mfocko: right… it's nested in :D but I'll include something, I did the same for the packit.dev

PR #196, opened 2023-08-07
fix: use ‹⋄› instead of ‹<>›

mfocko:
‹<>› is interpreted as tag

**fixes the packit.dev build**

TODO for Friday: add an action for test build on both research and deployment docs…


lbarcziova (approved):
thanks!

> TODO for Friday: add an action for test build on both research and deployment docs…

+1, that would be super useful, feel free to open issue so that it is not on you only

PR #198, opened 2023-08-25
Research workflow engines

lbarcziova:
Fixes https://github.com/packit/packit-service/issues/2033


research/workflow-engines/argo-workflows.md
```diff
@@ -0,0 +1,63 @@
...
+      container:
+        image: docker/whalesay
+        command: [cowsay]
+        args: ["{{inputs.parameters.message}}"]
+```
+
+- provides UI
+- workflows can be run using `Argo CLI` or directly `kubectl`
+- complex architecture: https://github.com/argoproj/argo-workflows/blob/master/docs/architecture.md
+- doesn't look like a fit for us
```
jpopelka, line 63: What makes you think that?

research/workflow-engines/index.md
```diff
@@ -0,0 +1,27 @@
...
+
+  - [Airflow](airflow.md)
+  - [Argo workflows](argo-workflows.md)
+  - [Temporal](temporal.md)
+  - [Selinon](selinon.md)
+  - [Prefect](prefect.md)
+  - [(Celery Canvas)](celery-canvas.md)
+
+- besides that, I found some great articles and resources overviewing some of these:
+  - overview and comparison of Argo, Airflow and Prefect: https://neptune.ai/blog/argo-vs-airflow-vs-prefect-differences
```
jpopelka, line 19: nice one

research/workflow-engines/prefect.md
```diff
@@ -0,0 +1,42 @@
+---
+title: Prefect
+authors: lbarcziova
+---
+
+- docs: https://docs.prefect.io/2.11.5/
+
+  > Prefect is a workflow orchestration tool empowering developers to build, observe, and react to data pipelines.
+
+- provides UI for monitoring flows and tasks, and it also allows you to build custom dashboards: ![UI](https://docs.prefect.io/2.11.5/img/ui/dashboard-cloud.png)
```
TomasTomecek, line 10: the UI is neat

TomasTomecek (approved):
Fascinating. I am curious about the outcome & next steps.

research/workflow-engines/argo-workflows.md
lbarcziova: the way of defining the workflows as templates, which seems too static and not fitting to how we would like to schedule the jobs and dynamically specify the dependencies between them

research/workflow-engines/airflow.md
```diff
@@ -0,0 +1,72 @@
...
+    # Set task dependencies
+    build >> test
+
+    return dag
+
+dynamic_workflow()
+
+```
+
+- Airflow uses schedulers and executors to manage task execution. The scheduler determines when to execute tasks, while the executor defines how tasks are run. Airflow supports various executors, including SequentialExecutor, LocalExecutor, and CeleryExecutor.
```
lachmanfrantisek, line 54: Do you know how we can make Airflow tasks to be triggered by an external event?

Will this require implementing our tasks in a different way to combine both starting and waiting for the result?

research/workflow-engines/airflow.md
```diff
@@ -0,0 +1,72 @@
...
+- allows dynamic pipelines generation
+- provides UI ![UI](https://airflow.apache.org/docs/apache-airflow/stable/_images/dags.png)
+- from `Why not` in docs:
+  > Airflow™ was built for finite batch workflows. While the CLI and REST API do allow triggering workflows, Airflow was not built for infinitely running event-based workflows. Airflow is not a streaming solution. However, a streaming system such as Apache Kafka is often seen working together with Apache Airflow. Kafka can be used for ingestion and processing in real-time, event data is written to a storage location, and Airflow periodically starts a workflow processing a batch of data.
+
+## CeleryExecutor
+
+Details: https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/executor/celery.html
+
+- it is also possible to use Flower UI
```
lachmanfrantisek, line 64: This looks interesting, have you had time to investigate how would this work? (What is the benefit of using Celery as an executor and if/how we can transition from pure Celery to Airflow+Celery?)

research/workflow-engines/argo-workflows.md
```diff
@@ -0,0 +1,63 @@
+---
+title: Argo Workflows
+authors: lbarcziova
+---
+
+- docs: https://argoproj.github.io/argo-workflows/
+
+  > Argo Workflows is an open source container-native workflow engine for orchestrating parallel jobs on Kubernetes. Argo Workflows is implemented as a Kubernetes CRD (Custom Resource Definition).
+  >
+  > - Define workflows where each step in the workflow is a container.
```
lachmanfrantisek, line 10: That looks like a really scalable solution. I am wondering if we can't use user-defined images for the sandbox. (Actually, that was the way [Betka](https://github.com/sclorg/betka) (the propose-downstream for containers) was designed.)

research/workflow-engines/argo-workflows.md
```diff
@@ -0,0 +1,63 @@
...
+
+- example of multi-step workflow config:
+
+```yaml
+apiVersion: argoproj.io/v1alpha1
+kind: Workflow
+metadata:
+  generateName: steps-
+spec:
+  entrypoint: hello-hello-hello
```
lachmanfrantisek, line 22: Could you please try to transform this example into our use cases so we can better imagine what this would look like?

(I really like how you've shown it on Airflow.)

research/workflow-engines/celery-canvas.md
```diff
@@ -0,0 +1,33 @@
...
+s = task_a.s(arg1, arg2)
+s.link(task_b.s(arg4, arg5))
+
+# another way using pipes
+work_flow = task_a(arg1, arg2) | task_b(result, arg4) | task_c(arg5, arg6)
+
+```
+
+- we would only write code on top of existing one, for handling the dependant tasks, no need
+  for deployment changes
```
lachmanfrantisek, line 32: So it can replace the decorators covering the job dependencies?

Do you think we can dynamically create this based on the user config (e.g. `skip_build` or an explicit user-defined chain in the future)?

research/workflow-engines/celery-canvas.md
```diff
@@ -0,0 +1,33 @@
...
+authors: lbarcziova
+---
+
+- docs: https://docs.celeryq.dev/en/stable/userguide/canvas.html
+- provides capabilities to build simple workflows
+- it is possible to define e.g. serial execution, in parallel, in group
+
+## Chains
+
+- tasks can be linked together: the linked task is called when the task returns successfully
```
lachmanfrantisek, line 12: Do you know if it's possible to include external events in this? (That's the key feature I am worried about across all the solutions -- how to include result reporting.)

research/workflow-engines/index.md
```diff
@@ -0,0 +1,27 @@
+---
+title: Workflow engines
+authors: lbarcziova
+---
+
+- this research aims to describe workflow engines that could improve the state of the service, in particular
+  how the logic around job dependencies could be eased by usage of these
+
+- you can find separate files describing tools that were proposed to be researched more:
```
lachmanfrantisek, line 9: Since you've spent a significant time on that, could you please add a conclusion with your opinion and suggest the next steps?

research/workflow-engines/selinon.md
```diff
@@ -0,0 +1,34 @@
...
+authors: lbarcziova
+---
+
+- docs: https://selinon.readthedocs.io
+
+> This tool is an implementation above Celery that enables you to define flows and dependencies in flows, schedule tasks based on results of Celery workers, their success or any external events.
+>
+> The main advantage of using Selinon over these are facts, that you can use it in fully distributed systems and for example let Kubernetes or OpenShift do the workload (and much more, such as recursive flows, sub-flows support, selective task runs, …), define task flows that are not strictly DAG (not strictly directed acyclic graph, but directed graph instead), let system compute which tasks should be run in dependency graph and other features.
+
+- migration from raw Celery: https://selinon.readthedocs.io/en/latest/celery.html
```
lachmanfrantisek, line 12: That's a clear plus!

lachmanfrantisek (commented):
Very nice overview!

I am yet to go through all the links.

research/workflow-engines/celery-canvas.md
lbarcziova: not sure how would you expect this to work :thinking: it is needed to create the tasks so that you will can link them together/specify the dependencies

research/workflow-engines/airflow.md
lbarcziova: for us, the clear benefit would be using a tool that we already know, the configuration, deployment around it etc.

but as for the transition, this way, Celery would be used for executing the tasks, but the Airflow schedulers would handle the logic of scheduling them, so the way tasks are handled would be changed.

there are multiple types of executors that can be used, see above, for the remote execution of tasks, besides Celery, one can also use KubernetesExecutor, CeleryKubernetesExecutor, etc. (I will add this here so that it is more explicit)

research/workflow-engines/argo-workflows.md
lbarcziova: I did not see a benefit in showing this here as this is a static yaml configuration and doesn't fit the nature of dynamically creating tasks that we do

but I will check how some client library could be used for dynamically scheduling the tasks and extend this

research/workflow-engines/index.md
lbarcziova: sure! I firstly wanted to gather some feedback on what to extend and then in the end provide the conclusion, I will add this today

research/workflow-engines/airflow.md
lachmanfrantisek: Maybe a `KubernetesExecutor` can be used for/intead-of the Sandcastle.

research/workflow-engines/celery-canvas.md
lachmanfrantisek: My point is that we don't have a workflow where everything happens on our side and the workflow engine "just" decides order and go one-by-one. Each step of our workflow is triggered on our side but needs results from an external service => and I am unsure how these engines can cover this. (The process looks like a workflow for the user, but for us, it's just one task done as a reaction to an external event.)

research/workflow-engines/index.md
lachmanfrantisek: Sure, thanks a lot!

research/workflow-engines/argo-workflows.md
```diff
@@ -0,0 +1,63 @@
+---
+title: Argo Workflows
+authors: lbarcziova
+---
+
+- docs: https://argoproj.github.io/argo-workflows/
+
+  > Argo Workflows is an open source container-native workflow engine for orchestrating parallel jobs on Kubernetes. Argo Workflows is implemented as a Kubernetes CRD (Custom Resource Definition).
```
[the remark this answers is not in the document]
lbarcziova: good point, I will extend this

research/workflow-engines/celery-canvas.md
lbarcziova: yes, I get that. But I am afraid this "custom" logic needs to be always implemented by us, no tool can do this for us :/ But I will give it more thought.

research/workflow-engines/airflow.md
TomasTomecek: I'm sure that over the years, a solution better than sandcastle has to be out there. The idea of sandcastle is simple and we're not the only one that need that.

Anyway, I found https://github.com/confidential-containers/confidential-containers/blob/main/quickstart.md

which look promising, but under the hood, it's kata containers ( = VMs) that use [TEE](https://en.wikipedia.org/wiki/Trusted_execution_environment)s: this solution is far more expensive than sandcastle

Red Hat's docs: https://access.redhat.com/documentation/en-us/openshift_sandboxed_containers/1.4/html-single/openshift_sandboxed_containers_user_guide/index#deploying-sandboxed-container-workloads-peer-pods

research/workflow-engines/celery-canvas.md
lbarcziova: actually, in Airflow there are "Sensors" that could probably help us with that, will add this

lachmanfrantisek: @lbarcziova , actually, @majamassarini just found that Temporal might have something for that and covered that in her research (#204)

PR #199, opened 2023-08-28
ci: fix build

mfocko:
That was not as good suggestion as it looked like :/


majamassarini (approved):
Thanks!!! I did not realize it was failing 🙈

mfocko:
It wasn't failing here, sorry :D

PR #204, opened 2023-09-29
Job referencing research

majamassarini:
Fix packit/packit-service#1773

### DIFF
--- a/research/job_referencing/index.md
+++ b/research/job_referencing/index.md
@@ -1,0 +1,239 @@
+---
+title: Job referencing
+authors:
+  - mmassari
+sidebar_position: 1
+---
+
+# Job referencing
+
+## Already implemented use cases
+
+The following dependencies use cases are not exposed to the user.
+This are implementation details useful as a starting point for thinking about user defined dependencies.
+
+### COPR build depends on SRMP build
+
+[Here](https://github.com/packit/packit-service/blob/16236087fb8071c83b3802464f3af504f4fb933e/packit_service/worker/handlers/copr.py#L136-L203) we build both the SRPM and the COPR packages (and the COPR package depends on the SRPM one).
+We don't need any special data to be saved in the database, we just check for the existance of the SRPM build.
+This code is triggered by a `CoprBuildStartEvent` event, this event is sent by COPR both for the SRPM package build and for the COPR package build, in this way we run twice the code and we can deal with the dependency.
+
+### Tests depends on COPR builds
+
+[Here](https://github.com/packit/packit-service/blob/16236087fb8071c83b3802464f3af504f4fb933e/packit_service/worker/handlers/testing_farm.py#L202-L241) we check that a COPR build exist for every target we should run a test for.
+If not build is found we run a `CoprBuldHandler` instance.
+For every `CoprBuildEndEvent` the `CoprBuildEndHandler` is able to run again the `TestingFarmHandler` if needed.
+No special data is stored in the database, the COPR build is searched using: project name, commit sha, copr build owner and target.
+
+## New use cases
+
+### Tests depends on specified COPR builds
+
+Since it is possible to add identifiers for jobs, we allow defining multiple Copr build jobs and multiple TF jobs. In that case, Packit Service doesn't know what Copr build job to use for the particular TF job.
+
+For this purpose, we could introduce a new field for job configs which would allow the referencing, for example:
+
+```yaml
+- job: copr_build
+identifier: "build1"
+trigger: pull_request
+actions: ...
+- job: copr_build
+identifier: "build2"
+trigger: pull_request
+actions: ...
+- job: tests
+identifier: "tests2"
+trigger: pull_request
+build_identifier: "build1"
+- job: tests
+identifier: "tests2"
+trigger: pull_request
+build_identifier: "build2"
+```
+
+#### The easiest solution I see
+
+I don't think we need anything more in the database to be able to retrieve the copr build id. We need the data given in the `copr_build` job matching the specified `build_identifier` name in the `tests` job, plus all the data we are already using in the search like the `commit_sha`.
+We need to add the `build_identifier` key in config.
+We need to modify [this code](https://github.com/packit/packit-service/blob/16236087fb8071c83b3802464f3af504f4fb933e/packit_service/worker/handlers/testing_farm.py#L202-L241) in a way that it is able to retrieve the proper `build_id`.
+
+#### `build_identifier` key name or?
+
+Another suggestion was to use `after` as a name for the key. I like it but our implementation, if made as specified above, will not be generic. For this reason I would rather prefer `after_build` or `build_identifier` to mark the fact that in the test job we are waiting for a build job.
+
+### Monorepo: Package dependencies
+
+The following is the [RFE](https://github.com/packit/packit/issues/1903) by LecrisUT
+
+```
+packages:
+A:
+depends:
+- B
+- name: C
+rebuild: true
+B: ...
+C: ...
+```
+
+at first I would simplify it like:
+
+```
+packages:
+A:
+depends:
+- B
+- C
+B: ...
+C: ...
+```
+
+I would always rebuild everything for every package but we can enhance the code and if a package has no changes in the given `commit_sha` and it does not depend on changed packages, we can simply copy lines from the previous database pipeline, filling the database as if the jobs were run.
+
+#### The easiest solution I see following what we are already doing for copr builds and tests
+
+We should [create tasks](https://github.com/packit/packit-service/blob/16236087fb8071c83b3802464f3af504f4fb933e/packit_service/worker/jobs.py#L464-L497) only for those packages that do not depend on other packages.
+When running, as an example, the COPR build for the latest dependent package in the list (C in our example), at the end of the handler we can create the COPR build tasks for package A (if any).
+This can be done for all the kinds of job we directly start at [this lines](https://github.com/packit/packit-service/blob/16236087fb8071c83b3802464f3af504f4fb933e/packit_service/worker/jobs.py#L464-L497).
+
+We can change `bodhi` jobs in a way that they will do nothing, apart from starting new `bodhi` tasks for other packages, unless they belong to a package not required by another package. Ideally only package A, in this example, should do a `bodhi` update if I am not wrong. Or at least we should be able to skip doing something in this job during the chain of dependencies.
+
+I think we don't need changes in the database schema to be able to implement this solution.
+We need to add the `depends` or `depends_on` key to the schema. I slightly prefer the `depends` key, but not strong opinions on that.
+
+### Request new feature: support side tag for multi package update
+
+This is the [RFE](https://github.com/packit/packit/issues/1870).
+
+I am not really sure I am getting the point here. But they are talking about a multi-package feature and for this reason I assume they need to use the monorepo syntax (otherwise we have no way to reference the packages).
+If they can and want to use the monorepo syntax then I think that the package dependencies solution for the monorepos would solve their problems too. Or am I wrong?
+
+## Drawbacks
+
+I see just a drawback in the previous solutions, the code is getting more and more complex to read (I mean to grasp the execution order) and test.
+I think `Celery` was well suited for tasks which didn't depend on each other, now that dependencies are growing quickly I feel it is like a limit. I would like to be able to have communication between tasks and suspend the execution without exiting.
+I think that a code using `asyncio`, as an example, could be more readable and testable.
+
+Let's take as an example the simple dependency we have between COPR builds and SRPM builds
+
+[This code](https://github.com/packit/packit-service/blob/16236087fb8071c83b3802464f3af504f4fb933e/packit_service/worker/handlers/copr.py#L136-L203) is performing both builds. But it is not straightforward to know that this code is called twice, because there are two `CoprBuildStartEvent`(s) sent by the COPR server. The execution flow is not obvious and not easily testable, even though this is a really simple example.
+
+I would like to be able to write something similar using `asyncio`.
+As far as I can understand it [temporalio](https://docs.temporal.io) would give us this flexibility. For this reason I wrote down some _python pseudo code_ using `temporalio`. I don't really know it so I can be missing something.
+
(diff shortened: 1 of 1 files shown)


research/job_referencing/index.md
```diff
@@ -0,0 +1,217 @@
...
+This can be done for all the kinds of job we directly start at [this lines](https://github.com/packit/packit-service/blob/16236087fb8071c83b3802464f3af504f4fb933e/packit_service/worker/jobs.py#L464-L497).
+
+We can change `bodhi` jobs in a way that they will do nothing, apart from starting new `bodhi` tasks for other packages, unless they belong to a package not required by another package. Ideally only package A, in this example, should do a `bodhi` update if I am not wrong. Or at least we should be able to skip doing something in this job during the chain of dependencies.
+
+I think we don't need changes in the database schema to be able to implement this solution.
+We need to add the `depends` or `depends_on` key to the schema. I slightly prefer the `depends` key, but not strong opinions on that.
+
+### Request new feature: support side tag for multi package update
+
+This is the [RFE](https://github.com/packit/packit/issues/1870).
```
lachmanfrantisek, line 104: I think the key ask is to be able to create an update including builds of multiple packages => we need to react when the last one is available and there will only be a single Bodhi update (per Fedora version).

Sadly, the monorepo way we have does not work here => we need one update for all the packages, not one for each.
A few ideas:
* We can specify the packages to include in a new config option and run the job only the one package (the last to be built, needs to be specified).
* We can run the job for all packages but only the last to be built will actually create the update, others will not do anything. (Seems a bit hackish to me.)

research/job_referencing/index.md
```diff
@@ -0,0 +1,217 @@
...
+
+I am not really sure I am getting the point here. But they are talking about a multi-package feature and for this reason I assume they need to use the monorepo syntax (otherwise we have no way to reference the packages).
+If they can and want to use the monorepo syntax then I think that the package dependencies solution for the monorepos would solve their problems too. Or am I wrong?
+
+## Drawbacks
+
+I see just a drawback in the previous solutions, the code is getting more and more complex to read (I mean to grasp the execution order) and test.
```
lachmanfrantisek, line 111: :+1:

research/job_referencing/index.md
```diff
@@ -0,0 +1,217 @@
...
+I think `Celery` was well suited for tasks which didn't depend on each other, now that dependencies are growing quickly I feel it is like a limit. I would like to be able to have communication between tasks and suspend the execution without exiting.
+I think that a code using `asyncio`, as an example, could be more readable and testable.
```
lachmanfrantisek, line 113: Do you mean to add some asynchronous waiting to have a single task (~code perspective) for a single job (~user perspective)?

research/job_referencing/index.md
```diff
@@ -0,0 +1,217 @@
...
+    await workflow.wait_condition(lambda: self._copr_builded)
+
+    update_ux_handle = workflow.start_activity(
+      update_ux,
+      start_to_close_timeout=timedelta(seconds=1),
+      ...
+      )
+```
+
+As far as I understand from the `temporalio` documentation, the framework is in charge of putting this `CoprBuildHandler workflow` in a sleeping queue while it is waiting for a signal and waking up it later. So we don't need to exit the workflow and the order of the activities is well visible in the `workflow` code.
```
lachmanfrantisek, line 212: This sounds awesome and exactly what we need!

research/job_referencing/index.md
```diff
@@ -0,0 +1,217 @@
...
+
+There is another small improvement I see using a solution like this one instead of doing what we are already doing.
+I don't think that the `CoprBuildEndHandler` should deal with the `TestingFarmHandler` and viceversa, this is dangerous, and the more dependencies we will introduce the more we will need this dangerous cross referencing.
```
lachmanfrantisek, line 215: Yes.

Ideally, I would leave the dependency definition to users and avoid as much magic/guessing as possible.

research/job_referencing/index.md
```diff
@@ -0,0 +1,217 @@
...
+
+Something above them should be able to orchestrate and know the jobs, and I think a `workflow` is exactly what we are missing here.
```
lachmanfrantisek, line 217: We can maybe discuss this before spending more time on that, but do you have an idea how we should transition to this (and how easy/hard this might be)?

research/job_referencing/index.md
```diff
@@ -0,0 +1,217 @@
...
+I think `Celery` was well suited for tasks which didn't depend on each other, now that dependencies are growing quickly I feel it is like a limit. I would like to be able to have communication between tasks and suspend the execution without exiting.
+I think that a code using `asyncio`, as an example, could be more readable and testable.
+
+Let's take as an example the simple dependency we have between COPR builds and SRPM builds
+
+[This code](https://github.com/packit/packit-service/blob/16236087fb8071c83b3802464f3af504f4fb933e/packit_service/worker/handlers/copr.py#L136-L203) is performing both builds. But it is not straightforward to know that this code is called twice, because there are two `CoprBuildStartEvent`(s) sent by the COPR server. The execution flow is not obvious and not easily testable, even though this is a really simple example.
+
+I would like to be able to write something similar using `asyncio`.
+As far as I can understand it [temporalio](https://docs.temporal.io) would give us this flexibility. For this reason I wrote down some _python pseudo code_ using `temporalio`. I don't really know it so I can be missing something.
```
lachmanfrantisek, line 120: (Just asking.) Have you considered any other solution or is this the only one the might be suitable for us?

lachmanfrantisek (commented):
Thanks a lot, Maja!

research/job_referencing/index.md
lachmanfrantisek: Thinking about that more, what about trying to use this e.g. for Copr builds (including all the steps) and run it inside the current celery task?

Or try this with something else to see how this can work in our use cases.

lachmanfrantisek: I am starting to like this framework. This is actually quite close to what we need: https://learn.temporal.io/tutorials/python/subscriptions/

research/job_referencing/index.md
```diff
@@ -0,0 +1,217 @@
...
+This code is triggered by a `CoprBuildStartEvent` event, this event is sent by COPR both for the SRPM package build and for the COPR package build, in this way we run twice the code and we can deal with the dependency.
+
+### Tests depends on COPR builds
+
+[Here](https://github.com/packit/packit-service/blob/16236087fb8071c83b3802464f3af504f4fb933e/packit_service/worker/handlers/testing_farm.py#L202-L241) we check that a COPR build exist for every target we should run a test for.
+If not build is found we run a `CoprBuldHandler` instance.
+For every `CoprBuildEndEvent` the `CoprBuildEndHandler` is able to run again the `TestingFarmHandler` if needed.
+No special data is stored in the database, the COPR build is searched using: project name, commit sha, copr build owner and target.
+
+## New use cases
```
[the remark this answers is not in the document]
majamassarini: I think we can do this in the same way we can deal with the `depends on package` in monorepo.
We should avoid running a Task for those tests which are dependent on other tests and start the Task just when the latest tests in the dependency list have successfully run.
After the changes I see our `TestingFarmHandler` will be able to run both new `CoprBuildHandlers` and new `TestingFarmHandlers` at its end.

research/job_referencing/index.md
```diff
@@ -0,0 +1,217 @@
...
+trigger: pull_request
+actions: ...
+- job: tests
+identifier: "tests2"
+trigger: pull_request
+build_identifier: "build1"
+- job: tests
+identifier: "tests2"
+trigger: pull_request
+build_identifier: "build2"
```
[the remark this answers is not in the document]
majamassarini: I don't know, but I would wait for something like `n x m`.  I don't feel like we have the right tools to work with dependencies. So for the moment I will keep the code as simple as we can, but I agree that we should satisfy our user's requests.
Probably the `1 x n` use case is not much more complicated than the `1 x 1` use case and it rassemble the `depends on package` for monorepo use case. So I think we can do it in a second step.

research/job_referencing/index.md
```diff
@@ -0,0 +1,217 @@
...
+
+#### The easiest solution I see
+
+I don't think we need anything more in the database to be able to retrieve the copr build id. We need the data given in the `copr_build` job matching the specified `build_identifier` name in the `tests` job, plus all the data we are already using in the search like the `commit_sha`.
+We need to add the `build_identifier` key in config.
+We need to modify [this code](https://github.com/packit/packit-service/blob/16236087fb8071c83b3802464f3af504f4fb933e/packit_service/worker/handlers/testing_farm.py#L202-L241) in a way that it is able to retrieve the proper `build_id`.
+
+#### `build_identifier` key name or?
+
+Another suggestion was to use `after` as a name for the key. I like it but our implementation, if made as specified above, will not be generic. For this reason I would rather prefer `after_build` or `build_identifier` to mark the fact that in the test job we are waiting for a build job.
```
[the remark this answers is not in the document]
majamassarini: This to me means something like being able to start from any handler any other handler in a generic way. I don't like the idea at the moment, maybe after a refactoring and using new tools this can become easier but at the moment I think this is risky.

research/job_referencing/index.md
```diff
@@ -0,0 +1,217 @@
...
+packages:
+A:
+depends:
+- B
+- C
+B: ...
+C: ...
+```
+
+I would always rebuild everything for every package but we can enhance the code and if a package has no changes in the given `commit_sha` and it does not depend on changed packages, we can simply copy lines from the previous database pipeline, filling the database as if the jobs were run.
```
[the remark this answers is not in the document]
majamassarini: Yes I agree that we need the filter out code.

research/job_referencing/index.md
majamassarini: I thought this could solve the problem:

>We can change `bodhi` jobs in a way that they will do nothing, apart from 
 starting new `bodhi` tasks for other packages, unless they belong to a 
 package not required by another package. Ideally only package A, in this 
 example, should do a `bodhi` update if I am not wrong. Or at least we should 
 be able to skip doing something in this job during the chain of dependencies.

research/job_referencing/index.md
majamassarini: At least for the more complex jobs like the copr build that depends on a srpm or the test job that depends on a copr build.

lachmanfrantisek: Thanks, I should have looked at the text and example bellow before commenting..;)

research/job_referencing/index.md
majamassarini: > Thinking about that more, what about trying to use this e.g. for Copr builds (including all the steps) and run it inside the current celery task?
> 
> Or try this with something else to see how this can work in our use cases.

I will think more about it. I don't think this can be done just using `asyncio`; we will end up, I think, having a lot of long-running workers and I think this is not ideal. But maybe I am wrong, I don't really know how Celery copes with asyncio.

research/job_referencing/index.md
majamassarini: I went through all the workflow engines listed in the research, from my point of view temporal.io is the only one that will allow us to write *linear code* and, if I get it right, it will take the burden to manage the sleeping tasks in a way completely transparent to us. But I may be wrong, I don't know any of them.

research/job_referencing/index.md
lachmanfrantisek: You are right, this combination might not be doable.

research/job_referencing/index.md
lachmanfrantisek: Thanks!

research/job_referencing/index.md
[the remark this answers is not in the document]
lachmanfrantisek: Yes, ideally, use some framework for that. I am worried resolving all the dependencies one by one might increase the code complexity a lot compared to a generic solution.

research/job_referencing/index.md
```diff
@@ -0,0 +1,239 @@
...
+    CON: it does not scale well, every new reference makes our code less readable/testeable and makes managing errors harder
+
+2.  Introduce a workflow engine with support for tasks communication
+
+    PRO: it should make our code more readable/testeable and we should be able to better manage error handling
+
+    CON: it will take us far more time
+
+An hybrid solution could be: go with solution 1 and implement just those job referencing we believe are actually/really needed.
+If we realize we would need other job referencing in future, we want to make job referencing more generic or we want the user to be able to build his own workflow then we need to start studying a workflow manager (like temporal) and planning a change in our core code base to go with solution 2.
```
mfocko, line 239: I would probably keep in mind the need to deprecate whatever comes out of the first solution.
Yes, we can adjust the configs almost automatically, but still.

research/job_referencing/index.md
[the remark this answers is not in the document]
lbarcziova: agreed, the code handling the current dependencies (e.g. build-> TF) is already making the code complexity much worse

research/job_referencing/index.md
```diff
@@ -0,0 +1,239 @@
...
+
+## Drawbacks
+
+I see just a drawback in the previous solutions, the code is getting more and more complex to read (I mean to grasp the execution order) and test.
+I think `Celery` was well suited for tasks which didn't depend on each other, now that dependencies are growing quickly I feel it is like a limit. I would like to be able to have communication between tasks and suspend the execution without exiting.
+I think that a code using `asyncio`, as an example, could be more readable and testable.
+
+Let's take as an example the simple dependency we have between COPR builds and SRPM builds
+
+[This code](https://github.com/packit/packit-service/blob/16236087fb8071c83b3802464f3af504f4fb933e/packit_service/worker/handlers/copr.py#L136-L203) is performing both builds. But it is not straightforward to know that this code is called twice, because there are two `CoprBuildStartEvent`(s) sent by the COPR server. The execution flow is not obvious and not easily testable, even though this is a really simple example.
```
lbarcziova, line 120: agreed :/

research/job_referencing/index.md
lbarcziova: +1

lbarcziova (approved):
thanks Maja, nicelly done!

PR #211, opened 2024-01-31
Add script to automate Fedora release onboarding

lbarcziova:
The script can generate the Packit config for dist-git repo and push it to rawhide or create a PR against rawhide with the config.

Fixes packit/packit-service#2301


research/fedora-packages-onboarding/onboard_package.py
```diff
@@ -0,0 +1,357 @@
...
+        value = locals().get(key)
+        if value:
+            config[key] = value
+
+    config["jobs"] = [
+        {
+            "job": "pull_from_upstream",
+            "trigger": "release",
+            # TODO also make this as input parameter?
+            "dist_git_branches": ["fedora-all"],
```
lbarcziova, line 86: I am not sure here if we got this from input, whether these should be 3 separate arguments then (for each job), or one argument (pull_from_upstream and koji_build would use the input value directly and for bodhi-update we would compute the {targets-from-input} - {rawhide}

lbarcziova (commented):
opening this PR for some early reviews and feedback

I was thinking about where to put this script (somewhere separately as in here, or create this as a Packit command) and am open for feedback/suggestions, because I see this could be used both by
- maintainers who already used Packit and want to automatically generate the config for lot of packages (we could build upon this script with reading the input values from some JSON or similar)
- for us (Packit devs) if we try to onboard some package to ease our work

research/fedora-packages-onboarding/onboard_package.py
lbarcziova: thanks for the feedback, agreed!

PR #212, opened 2024-05-15
Add build reverse dependencies research

majamassarini:
Fix (partially) packit/packit-service#2361

### DIFF
--- a/research/build-reverse-dependencies/index.md
+++ b/research/build-reverse-dependencies/index.md
@@ -1,0 +1,98 @@
+---
+title: Build reverse dependencies in Packit
+authors: mmassari
+---
+
+Already existing solutions:
+
+- [Koschei service](https://github.com/fedora-infra/koschei)
+- [Mass Prebuild CLI](https://gitlab.com/fedora/packager-tools/mass-prebuild)
+- [mini mass rebuild](https://github.com/hroncok/mini-mass-rebuild)
+- [copr-autorebuilder](https://pagure.io/copr-autorebuilder/tree/master)
+- [copr-rebuild-tools](https://github.com/fedora-copr/copr-rebuild-tools/tree/main)
+
+The first two in the above list are specifically designed for resolving reverse dependencies and this research will focus on them.
+
+## Koschei
+
+[Koschei](https://apps.fedoraproject.org/koschei) is a **continuous integration service** for RPM packages that tracks package dependency changes in Fedora Rawhide and rebuilds packages whose dependencies change too much.
+
+**Koschei depends on Koji**. New Koji builds "trigger" Koschei and Koschei rebuilds reverse dependencies as Koji scratch builds.
+**Koji scratch builds do not trigger Koshei** at the moment.
+This means that Koschei can be used as a CI service only after a _downstream release synchronization_ but **our users would like to check changes before making a downstream release synchronization.**
+
+If we want to use Koshei as an upstream CI service (or even downstream but without the need of a downstream release synchronization) we need to **make it react to scratch builds**.
+
+The Koschei project is not much lively, at least upstream; different reported issues are not being answered or worked on. For this reason I fear integrating new features in the Koschei production service could take a long time.
+
+If we want to use Koschei we could also think about **fork the project** and create our own instance inside the Packit cluster. This approach could give us two advantages:
+
+- Shorter time to production and our changes can always be taken back by the original project at some point.
+- We can re-enable COPR as a backend for Koschei builds. Accordingly to [this thread](https://pagure.io/fedora-ci/general/issue/45) Koschei has had a COPR backend for a while, but it has been removed (because COPR is a third party service not supported by the Fedora Infrastructure). Enabling COPR as a backend will allow the service to work even for those upstream projects which are not completely integrated in Fedora (if needed).
+
+What are the **PROs** I see in re-using Koshei:
+
+- We could benefit from an already well known somehow "standardized" user interface (the Koschei frontend).
+- We shouldn't need to be able to calculate reverse dependencies in Packit and submit builds from Packit if we will reuse (the Koschei backend).
+
+What are the **CONs** I see in re-using Koshei:
+
+- As far as I can understand it, the guy who wrote Koshei does not work anymore in Red Hat. We can not ask him for any help dealing with the Koschei code.
+- We could face unknown problems.
+
+## Mass Prebuild
+
+This is a CLI tool that resolves reverse dependencies and rebuild them all. It can work with both Koji and COPR.
+Even if it is a CLI tool we can not easily integrate it in packit-service because of two main design choices:
+
+- the tool interacts with a database
+- the tool keeps a session and can interact with the user
+
+The above design choices make the command line tool and also the code behind it not reusable as it is.
+
+## Build our own solution
+
+We will need:
+
+1. a way to resolve reverse dependencies
+
+   - we can follow the [Koschei way](https://github.com/fedora-infra/koschei/blob/master/koschei/backend/depsolve.py) which uses `hawkey` from `libdnf`
+   - or we can follow the [Mass prebuild way](https://gitlab.com/fedora/packager-tools/mass-prebuild/-/blob/main/mass_prebuild/utils/whatrequires.py?ref_type=heads): given a package or a set of packages, namely "main packages", the mass pre-builder will calculate the list of its direct reverse dependencies: packages that explicitly mark one of the main packages in their "BuildRequires" field. As far as I can understand it, this is a **less deep** and thus **less complete** approach compared with the Koschei solution.
+   - we can **avoid calculating dependencies** and let the user specify them in a static way.
+
+2. a way to schedule builds in Koji or COPR (or both).
+
+   We are quite good at submitting builds in both services but we are not good at dealing with dependencies.
+
+   - **if we don't need to interact with the rebuild process** we can depend upon:
+
+     - the [COPR build batches](https://docs.pagure.org/copr.copr/user_documentation.html#build-batches) feature that allows you to define the order of your builds in advance.
+
+     ```
+           $ copr build <project> --no-wait <first.src.rpm>
+           Created builds: 101010
+           $ copr build <project> --no-wait <second.src.rpm> --after-build-id 101010
+           Created builds: 101020
+           $ copr build <project> --no-wait <third.src.rpm> --with-build-id 101020
+           Created builds: 101030
+     ```
+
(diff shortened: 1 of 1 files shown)


research/build-reverse-dependencies/index.md
```diff
@@ -0,0 +1,98 @@
+---
+title: Build reverse dependencies in Packit
+authors: mmassari
+---
```
LecrisUT, line 4: General comment, what are the project structures that would support such reverse dependencies. Firstly w.r.t. where the builds will live
- copr project with multiple packages
  - Easy to pick up the latest changes
  - Need to manually clean the project build, otherwise deleted packages will still be used (e.g. a package is already downstream and the old copr package is no longer needed)
  - Bootstrap packages would be weird to handle
- multiple copr projects linked via `additional_repos`
  - Maximum control of the packages
  - Messy to navigate
- koji based builds
  - Good control of packages to be used
  -  Not universal for upstream projects

LecrisUT, line 4: Second part of the general comment about the supported project structures. Where are the packages to be rebuilt

- Monorepo
  - Easy to control what triggers build chains
  - Probably already the discussion of a different research/issue
- Monorepo with external references, i.e. point a package to a git repo and use the `packit.yaml` inside there
  - Good control of what packages to rebuild
  - Requires dependent project to have packit integration
- Completely external reference
  - Source and build environment can be anything, dist-git, copr project or packit project
  - Can dynamically query all dependencies using `fedrq` (only in case of dist-git source)

research/build-reverse-dependencies/index.md
```diff
@@ -0,0 +1,98 @@
...
+The above design choices make the command line tool and also the code behind it not reusable as it is.
+
+## Build our own solution
+
+We will need:
+
+1. a way to resolve reverse dependencies
+
+   - we can follow the [Koschei way](https://github.com/fedora-infra/koschei/blob/master/koschei/backend/depsolve.py) which uses `hawkey` from `libdnf`
+   - or we can follow the [Mass prebuild way](https://gitlab.com/fedora/packager-tools/mass-prebuild/-/blob/main/mass_prebuild/utils/whatrequires.py?ref_type=heads): given a package or a set of packages, namely "main packages", the mass pre-builder will calculate the list of its direct reverse dependencies: packages that explicitly mark one of the main packages in their "BuildRequires" field. As far as I can understand it, this is a **less deep** and thus **less complete** approach compared with the Koschei solution.
```
LecrisUT, line 60: Related to option 2 here. I think `fedrq` would be a more manageable alternative. Not sure about managing copr or non Fedora dependencies though, but maybe the tool can evolve there.

research/build-reverse-dependencies/index.md
```diff
@@ -0,0 +1,98 @@
...
+   - we can **avoid calculating dependencies** and let the user specify them in a statistical way.
+
+2. a way to schedule builds in Koji or COPR (or both).
```
LecrisUT, line 63: This point 2 is needed more generally, e.g. for monorepo

research/build-reverse-dependencies/index.md
LecrisUT, line 4: (self-promotion) I was starting to implement this as a tmt-plugin ([tmt-copr](https://github.com/LecrisUT/tmt-copr) and also tmt-koji, but need to finish 1 first :D)
- Still requires testing-farm integration to pull in additional tmt plugins
- Less to manage on packit side
- More control on how to display results, logs
- Needs the secret management of testing-farm to get the api token to interact with services. But an alternative could incorporate packit cli and packit service to get things working faster

This can solve some of the issues of "Build our own solution", especially w.r.t. viewing the results

majamassarini: I don't think there is a need for changing the project structure but I am not sure I am following you.
The `reverse-dependencies` should be dynamically rebuilt (and also discovered if we don't go in the way of statically defining them).
This is done more easily when we work with Koji and the Fedora ecosystem where all the dependencies should already be satisfied, https://koschei.fedoraproject.org/ already does it and you don't need to specify anything.
If we decide to use COPR and enable every upstream repo we should probably find a way to describe in the new `Job` the reverse dependencies (I would say statically) and a way to build them in COPR (there could be many approaches for this also the `action` can be used).
For me this is a "last step", an enhancement in the new `Job` but if I understand it correctly you would benefit from it because not all your packages are released in Fedora?

research/build-reverse-dependencies/index.md
majamassarini: Thanks, I will git it a look.

research/build-reverse-dependencies/index.md
LecrisUT: Re: side-tag, that's for the case of making the reverse dependency build a production workflow right? Can scratch builds also have side-tag or equivalent?

majamassarini: > Can scratch builds also have side-tag or equivalent?
You can assign a side tag to a scratch build.

If for production workflow you mean you need to have your commit released in dist-git. This is already supported in Koshei. We are looking for a way to have the workflow triggered before a new release in dist-git.

LecrisUT: > but if I understand it correctly you would benefit from it because not all your packages are released in Fedora?

Yes, this is for pre-release/pre-packaging. Referencing a git-repo under packit would be sufficient for that.

> If we decide to use COPR and enable every upstream repo

The main supported upstream model here sould be to build in copr with dist-git source (probably from `fedrq`).

> I don't think there is a need for changing the project structure 

I was thinking of the structure of monorepo, just because it would be more natural to navigate a dependency tree and it would share much of the scheduling logic that is needed for monorepo dependencies.

Indeed an alternative would be to have it in Jobs, but I'm afraid complex dependencies would become harder to manage.

These would be another upstream model for when some dependencies are not yet packaged.

LecrisUT: I'm thinking w.r.t. the end result. Will it eventually be used in a bodhi update, or will it be used only for testing dependencies. The former might be needed only sporadically, like for SOVERSION bump, etc. but the latter could be used more freely. Also the former could be covered by the dist-git label workflow

research/build-reverse-dependencies/index.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Build our own solution
+
+We will need:
+
+1. a way to resolve reverse dependencies
+
+   - we can follow the [Koschei way](https://github.com/fedora-infra/koschei/blob/master/koschei/backend/depsolve.py) which uses `hawkey` from `libdnf`
+   - or we can follow the [Mass prebuild way](https://gitlab.com/fedora/packager-tools/mass-prebuild/-/blob/main/mass_prebuild/utils/whatrequires.py?ref_type=heads): given a package or a set of packages, namely "main packages", the mass pre-builder will calculate the list of its direct reverse dependencies: packages that explicitly mark one of the main packages in their "BuildRequires" field. As far as I can understand it, this is a **less deep** and thus **less complete** approach compared with the Koschei solution.
+   - we can **avoid calculating dependencies** and let the user specify them in a statistical way.
```
lbarcziova, line 61:
```suggestion
   - we can **avoid calculating dependencies** and let the user specify them in a static way.
```

research/build-reverse-dependencies/index.md
```diff
@@ -0,0 +1,98 @@
...
+authors: mmassari
+---
+
+Already existing solutions:
+
+- [Koschei service](https://github.com/fedora-infra/koschei)
+- [Mass Prebuild CLI](https://gitlab.com/fedora/packager-tools/mass-prebuild)
+- [mini mass rebuild](https://github.com/hroncok/mini-mass-rebuild)
+- [copr-autorebuilder](https://pagure.io/copr-autorebuilder/tree/master)
+- [copr-rebuild-tools](https://github.com/fedora-copr/copr-rebuild-tools/tree/main)
```
lbarcziova, line 12: I assume these would probably not be a good fit since they are not included, right?

research/build-reverse-dependencies/index.md
```diff
@@ -0,0 +1,98 @@
...
+- As far as I can understand it, the guy who wrote Koshei does not work anymore in Red Hat. We can not ask him for any help dealing with the Koschei code.
+- We could face unknown problems.
+
+## Mass Prebuild
+
+This is a CLI tool that resolves reverse dependencies and rebuild them all. It can work with both Koji and COPR.
+Even if it is a CLI tool we can not easily integrate it in packit-service because of two main design choices:
+
+- the tool interacts with a database
+- the tool keeps a session and can interact with the user
```
lbarcziova, line 49: so there is no way of disabling this interactive mode?

research/build-reverse-dependencies/index.md
```diff
@@ -0,0 +1,98 @@
...
+
+     - or/and the [Koji chained builds](https://docs.fedoraproject.org/en-US/package-maintainers/Using_the_Koji_Build_System/#chained_builds)
+
+     ```
+           fedpkg chain-build --target=side-tag-name libwidget libaselib : libgizmo :
+     ```
+
+   - but we should be aware that **the mass prebuild tool is designed in a way that lets you interact with every single step of the long rebuild process**. I don't know exactly why, but if I got it correctly, this process is error prone and the user may like to be able to continue it from a point of failure or may like to resubmit a failed build. I think it is worth discussing with the mass pre build developer why he has chosen to go this way.
+
+     If we want the user to be able to interact with every step in this long process we should probably be capable of dealing with **job dependencies**.
```
lbarcziova, line 88: I hope the more straightforward way, i.e. using batch builds/ chain builds without the need to solve the dependencies ourselves will be enough. We should probably get feedback from the potential users for this so that we can decide.

lbarcziova (commented):
thanks Maja, this is really nicely written down! :rocket:

We should definitely discuss within the team the options.

research/build-reverse-dependencies/index.md
majamassarini: Yes, they are *mass rebuild tools* not focused on *reverse dependencies*.
We can give them a further look if we decide to implement our own solution and we want to investigate more on how to submit a large number of builds.

research/build-reverse-dependencies/index.md
majamassarini: yes, you can decide to submit one step at time, specifying which step you want, but still the code is highly coupled with the "entire session idea" and the database.

research/build-reverse-dependencies/index.md
lbarcziova: makes sense, thanks!

research/build-reverse-dependencies/index.md
lbarcziova: understood :+1:

lbarcziova (approved):
:rocket:

PR #217, opened 2024-09-11
Add Fedora CI research

mfocko:
Fixes packit/packit-service#2462

### DIFF
--- a/research/integrations/fedora-ci/index.md
+++ b/research/integrations/fedora-ci/index.md
@@ -33,11 +33,40 @@ the results of the SRPM build, i.e., just run `fedpkg srpm` and be done with it.
 This can be done within the sandbox as we do for non-scratch builds. However it
 might need some tweaking as Packit has specific worfklow.
 
+:::note Questions from review
+
+> This is a very interesting point -- do we want to be consistent as much as
+> possible with Packit's upstream implementation or be close to the real Fedora
+> builds.
+>
+> I agree the suggested approach `fedpkg srpm` makes more sense from the user
+> perspective since this results in the production-like result.
+
+With respect to downstream, I would be definitely voting for avoiding adjusting
+the SRPM creation. Adjusting the SRPM creation in any way enables the chance of
+inconsistencies, i.e., I have green CI, but once I merge it doesn't build. I
+don't think many maintainers would like that :D
+
+> How do we want to handle sources? (Do we want to continue with the current
+> trend of requiring sources in the lookaside cache?)
+
+I'm not really sure, but if we go deep into the possible improvements then it's
+quite possible that most of this initiative will not be achievable :)
+
+:::
+
 ### Koji scratch build
 
 We are already submitting scratch builds for merged PRs, if configured. This
 would just extend the functionality to the non-merged PRs (with need to provide
 the SRPM).
+
+:::note Review
+
+And we have scratch build functionality for upstream PRs available as well if
+this is anyhow relevant.
+
+:::
 
 ### RPM linting, inspect and installability test
 
@@ -46,6 +75,18 @@ There are premade tmt plans that we can run.
 :::note Question
 
 Should these be run in parallel or sequentially and fail on first?
+
+Follow-up question from review:
+
+> Are some of these somehow dependent?
+
+From the excerpt here (linting, installability, tests), it feels like linting
+should not be a hard blocker. I've seen the installability tests fail (even if
+tests passed) just because of the network flakes and such, so…
+
+In the sense of saving the resources, it definitely makes sense to make _full_
+_test suite_ dependent on the _installability_, but if it's flakey, it's just
+annoying.
 
 :::
 
@@ -156,6 +197,29 @@ Most notable differences:
 Preferably it would be ideal to have separate deployment for the “dist-git only”
 service as handling all Fedora packages would definitely increase traffic and
 could cause issues with running as both Fedora CI and original Packit Service.
+
+:::note Question from review
+
+> Would this mean `dist-git CI` and the current Packit? Or, Packit upstream and
+> Packit Fedora automation (including Koji, Bodhi and maybe also propose/pull)?
+
+I've been thinking about a way to split this in some reasonable matter :D Based
+on the way Packit operates, the most reasonable would be splitting the
+upstream/downstream parts, i.e.,
+
+- Packit Service (handles upstream **and** `propose-downstream` (as it is
+  triggered by upstream))
+- “Packit as Fedora CI” (handles downstream **and** `pull-from-upstream`)
+
+**However** I rant somewhere about the config in the dist-git and default config
+:D and this complicates things a lot :/ Maybe we could treat PRs in a special
+way for the _downstream PRs_ (Koji + TF).
+
+---
+
+feels like whack-a-mole…
+
+:::
 
 #### Default config
 
@@ -279,9 +343,25 @@ workflows, considerable overhead.
 
 ### GitLab
 
-ogr has a support for GitLab already, reporting sucks a bit, but it's not
-a blocking issue (due to the lack of upstream users, the priority of improving
-the UX has been lowered).
+ogr has a support for GitLab already, reporting sucks, but it's not a blocking
+issue (due to the lack of upstream users, the priority of improving the UX has
+been lowered).
+
+:::warning GitLab UX
+
+There are limitations on how Packit can improve the user experience on GitLab.
+
+GitLab is focused on providing the best user experience when using their own
+GitLab CI, external services must either just report commit statuses (unlike on
+GitHub that allows Check Suites and Check Runs), or integrate into the GitLab
+codebase itself.
+
+This hinders the user experience significantly.
+
+> Switching from Pagure to GitLab sounds effectively like getting rid of the
+> stability issues while maintaining the poor UX.
+
+:::
 
 #### Upstream vs downstream
 
@@ -320,14 +400,17 @@ helpful in this case a lot).
 - [ ] Consider this in the design.
 - [ ] Should not cause any annoyance for _proven packagers_
 
+:::note Review
+
+There's an [RFE](https://github.com/packit/packit-service/issues/1723) for
+a workflow that involves auto-landing PRs in the Fedora dist-git.
+
+:::
+
 ## Follow-up cards
 
 - [ ] Allow triggering of the Testing Farm with Koji build (instead of the Copr
       build)
-- [ ] Allow triggering of the Testing Farm **STI** tests with Koji build (or in
-      general)
-- [ ] After PoC and final decision, deploy a separate instance to handle only
-      Fedora CI queue
 - [ ] Try to unify _installability pipeline_ with the tmt plan for install test
       that we already have
 - [ ] Packit needs to be able to deduce the build target based on the branch
@@ -336,3 +419,7 @@ helpful in this case a lot).
       from the config)
 - [ ] Allow Packit to have customizable default config (more details above)
 - [ ] Allow Packit to run without any config (utilizing the previous point)
+- [ ] After PoC and final decision, deploy a separate instance to handle only
+      Fedora CI queue
+- [ ] Allow triggering of the Testing Farm **STI** tests with Koji build (or in
+      general)
--- a/research/integrations/fedora-ci/index.md
+++ b/research/integrations/fedora-ci/index.md
@@ -1,0 +1,192 @@
+---
+title: Packit as Fedora dist-git CI
+authors: mfocko
+---
+
+Related links:
+
+- [Epic](https://github.com/packit/packit-service/issues/2453)
+- [Research issue](https://github.com/packit/packit-service/issues/2462)
+
+## Current Fedora CI
+
+1. Koji scratch build
+1. Installability
+1. tmt (if plans are present)
+
+### Zuul as a parallel CI
+
+1. Builts SRPM
+1. Submits Koji scratch build
+1. Runs RPM linter
+1. Runs RPM inspect
+1. Tests install
+1. Runs tmt
+
+## Possible implementation
+
+### SRPM build
+
+Ideally kept identical. Packit should not be involved in any way that can affect
+the results of the SRPM build, i.e., just run `fedpkg srpm` and be done with it.
+
+This can be done within the sandbox as we do for non-scratch builds. However it
+might need some tweaking as Packit has specific worfklow.
+
+### Koji scratch build
+
+We are already submitting scratch builds for merged PRs, if configured. This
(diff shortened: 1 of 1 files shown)


lbarcziova (approved):
thanks!

research/integrations/fedora-ci/index.md
```diff
@@ -146,6 +153,20 @@ Most notable differences:
...
+
+#### Default config
+
+Having a separate instance would also open up a door to a potential “default
+config” which could be different to our current one (SRPM, Copr build, TF) and
+more tailored to the needs of the Fedora CI.
+
+**TODO**:
+
+- [ ] Maybe have a default config configurable?
```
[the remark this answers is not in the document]
mfocko: IIRC, right now if there are no jobs configured on upstream, we have some implicit definitions (Copr build and TF).

The same could be done on downstream, but for Koji builds, and TF.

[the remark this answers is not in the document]
lbarcziova: I see now, could you please add this explanation there?

research/integrations/fedora-ci/index.md
```diff
@@ -0,0 +1,338 @@
...
+the results of the SRPM build, i.e., just run `fedpkg srpm` and be done with it.
+
+This can be done within the sandbox as we do for non-scratch builds. However it
+might need some tweaking as Packit has specific worfklow.
+
+### Koji scratch build
+
+We are already submitting scratch builds for merged PRs, if configured. This
+would just extend the functionality to the non-merged PRs (with need to provide
+the SRPM).
```
lachmanfrantisek, line 40: And we have scratch build functionality for upstream PRs available as well if this is anyhow relevant.

research/integrations/fedora-ci/index.md
```diff
@@ -0,0 +1,338 @@
...
+
+## Possible implementation
+
+### SRPM build
+
+Ideally kept identical. Packit should not be involved in any way that can affect
+the results of the SRPM build, i.e., just run `fedpkg srpm` and be done with it.
+
+This can be done within the sandbox as we do for non-scratch builds. However it
+might need some tweaking as Packit has specific worfklow.
```
lachmanfrantisek, line 34: This is a very interesting point -- do we want to be consistent as much as possible with Packit's upstream implementation or be close to the real Fedora builds.

I agree the suggested approach `fedpkg srpm` makes more sense from the user perspective since this results in the production-like result.

lachmanfrantisek: How do we want to handle sources? (Do we want to continue with the current trend of requiring sources in the lookaside cache?)

research/integrations/fedora-ci/index.md
```diff
@@ -0,0 +1,338 @@
...
+
+### RPM linting, inspect and installability test
+
+There are premade tmt plans that we can run.
+
+:::note Question
+
+Should these be run in parallel or sequentially and fail on first?
+
+:::
```
lachmanfrantisek, line 50: Are some of these somehow dependant?

research/integrations/fedora-ci/index.md
```diff
@@ -0,0 +1,338 @@
...
+  "run_time": 1207.608357,
+  "created": "2024-09-06T09:31:17.471815",
+  "updated": "2024-09-06T09:31:17.471844"
+}
+```
+
+Most notable differences:
+
+- RPMs are provided from Koji scratch build (following Fedora CI, which is
+  easier than self-hosting)
```
lachmanfrantisek, line 149: Do you want to reuse the FedraCI-triggered Koji scratch build or trigger our own to be independent?

research/integrations/fedora-ci/index.md
```diff
@@ -0,0 +1,338 @@
...
+- chroot is defined by the branch rather than config
+
+## Technical issues / questions
+
+### Load balancing
+
+Preferably it would be ideal to have separate deployment for the “dist-git only”
+service as handling all Fedora packages would definitely increase traffic and
+could cause issues with running as both Fedora CI and original Packit Service.
```
lachmanfrantisek, line 158: Would this mean `dist-git CI` and the current Packit? Or, Packit upstream and Packit Fedora automation (including Koji, Bodhi and maybe also propose/pull)?

research/integrations/fedora-ci/index.md
```diff
@@ -0,0 +1,338 @@
...
+      For context, we have [default jobs](https://packit.dev/docs/guide#jobs) on
+      the upstream in case there are no jobs specified. This could be leveraged
+      to provide some **reasonable** defaults for Koji builds and Testing Farm
+      runs on the downstream part.
+
+### SRPM build
+
+For Zuul is run via `fedpkg srpm`. Then it's provided for the Koji build.
+
+Packit should not interefere with the way the SRPM gets built.
```
lachmanfrantisek, line 178: :+1:

research/integrations/fedora-ci/index.md
```diff
@@ -0,0 +1,338 @@
...
+
+### Propagating Koji build
+
+As of now, the sources need to be already uploaded in the lookaside cache to run
+the RPM build in general (either Fedora CI or Zuul CI). Given that merges in
+dist-git are handled as fast-forward, it should be possible to just tag the
+build from the PR instead of rebuilding again (with the same sources and
+specfile).
```
lachmanfrantisek, line 186: That would be really awesome and close to the ?onGitLab approach. We might want to submit a Fedora change for this.

research/integrations/fedora-ci/index.md
```diff
@@ -0,0 +1,338 @@
...
+- [ ] Does it make sense to open a follow-up already, even if there's no PoC
+      yet?
+
+### RPM for Testing Farm
+
+Zuul provides its own (self-hosted) repository. Fedora CI just refers to the
+scratch build.
+
+Self-hosting the repository is not preferred, following the Fedora CI approach
+might be the best.
```
lachmanfrantisek, line 205: We can also reuse the Copr integration but being close to the final system might be better in this place.

research/integrations/fedora-ci/index.md
```diff
@@ -0,0 +1,338 @@
...
+### STI tests
+
+Older format of tests that can be run via Zuul. Based on the TF API, Testing
+Farm supports it too. Requesting STI tests from the Testing Farm should be
+possible, detection of their presence lies on Packit.
+
+**TODO**:
+
+- [ ] Should we support the STI tests?
+
```
lachmanfrantisek, line 216: We can ask TF if we shouldn't motivate people to go away by not supporting it...;)

lachmanfrantisek: Is the presence detection of STI hard?

research/integrations/fedora-ci/index.md
```diff
@@ -0,0 +1,338 @@
...
+### Side tags
+
+Support for side tags has been requested in the original thread with the
+proposal. This may be required for “packaging workflows” that are more complex,
+e.g., Rust packaging (dependencies that are built as separate packages).
+
+> As a maintainer, I should be able to specify the dependencies of my package
+> that need to be verified with any proposed update.
```
lachmanfrantisek, line 224: Yes, manually-defined dependencies or reverse-dependencies are what people have been interested in so far.

research/integrations/fedora-ci/index.md
```diff
@@ -0,0 +1,338 @@
...
+Additionally, there needs to be a way to distinguish between upstream/downstream
+jobs anyways, cause Testing Farm runs on the downstream install RPMs from the
+Koji rather than Copr.
+
+:::warning
+
+It would probably be easier to run stateless as the current CI, utilizing only
+defaults and ignoring any Packit config that is present.
+
+:::
```
lachmanfrantisek, line 241: So, use Packit config just as an enabler (for now)?

research/integrations/fedora-ci/index.md
```diff
@@ -0,0 +1,338 @@
...
+### Fedora SIGs
+
+AFAIK the repositories are hosted on the Fedora dist-git, therefore they should
+be handled by the Fedora CI already, the experience should be similar (if not
+the same) to the regular Fedora packages.
+
+### CentOS SIGs
+
+As there is no streamlined CI for the CentOS SIGs, adoption would allow
+relatively simple integration with the Testing Farm.
```
lachmanfrantisek, line 271: Thanks for covering this one as well. Would be really good to count with these from the start.

research/integrations/fedora-ci/index.md
```diff
@@ -0,0 +1,338 @@
...
+
+### Configuration
+
+For the beginning we should just handle the packages that are configured.
+
+:::warning
+
+This means that we don't want to handle _all_ Fedora packages for the _PoC/MVP_.
+
+:::
```
lachmanfrantisek, line 310: Yes, but count on this possibility  (or global config ala zuul) since this should be the ultimate state.

research/integrations/fedora-ci/index.md
```diff
@@ -0,0 +1,338 @@
...
+
+### Auto-merging / gating
+
+During the implementation, we should keep in mind the possibility of polishing
+the process to eventually allow for auto-merging with improved gating (TF is
+helpful in this case a lot).
```
lachmanfrantisek, line 316: Good point, you can link https://github.com/packit/packit-service/issues/1723

This also looks like another reason to have a separate instance.

research/integrations/fedora-ci/index.md
```diff
@@ -0,0 +1,338 @@
...
+
+**TODO**:
+
+- [ ] Consider this in the design.
+- [ ] Should not cause any annoyance for _proven packagers_
+
+## Follow-up cards
```
lachmanfrantisek, line 323: Are these ordered? I would maybe put STI to the end or lower at least.

lachmanfrantisek (commented):
Wow, you've included a lot of interesting subjects!

research/integrations/fedora-ci/index.md
mfocko: > Is the presence detection of STI hard?

Have no clue :) I have **never** even seen the STI tests in the dist-git :D

research/integrations/fedora-ci/index.md
mfocko: > This is a very interesting point -- do we want to be consistent as much as possible with Packit's upstream implementation or be close to the real Fedora builds.
> 
> I agree the suggested approach `fedpkg srpm` makes more sense from the user perspective since this results in the production-like result.

With respect to downstream, I would be definitely voting for avoiding adjusting the SRPM creation. Adjusting the SRPM creation in any way enables the chance of inconsistencies, i.e., I have green CI, but once I merge it doesn't build. I don't think many maintainers would like that :D

> How do we want to handle sources? (Do we want to continue with the current trend of requiring sources in the lookaside cache?)

I'm not really sure, but if we go deep into the possible improvements then it's quite possible that most of this initiative will not be achievable :)

research/integrations/fedora-ci/index.md
mfocko: I thought we do non-scratch on upstream :eyes: I don't know how Koji handles “the heat”

research/integrations/fedora-ci/index.md
mfocko: From the excerpt here (linting, installability, tests), it feels like linting should not be a hard blocker. I've seen the installability tests fail (even if tests passed) just because of the network flakes and such, so…

In the sense of saving the resources, it definitely makes sense to make _full test suite_ dependent on the _installability_, but if it's flakey, it's just annoying.

research/integrations/fedora-ci/index.md
mfocko: :thinking: If Packit is to replace Fedora CI, there is no _Fedora CI-triggered Koji scratch build_ :P

research/integrations/fedora-ci/index.md
mfocko: I've been thinking about a way to split this in some reasonable matter :D Based on the way Packit operates, the most reasonable would be splitting the upstream / downstream parts, i.e.,

- Packit Service (handles upstream **and** `propose-downstream` (as it is triggered by upstream))
- “Packit as Fedora CI” (handles downstream **and** `pull-from-upstream`)

**However** I rant somewhere about the config in the dist-git and default config :D and this complicates things a lot :/ Maybe we could treat PRs in a special way for the _downstream PRs_ (Koji + TF).

---

feels like whack-a-mole…

research/integrations/fedora-ci/index.md
mfocko: With regards to your question about handling sources (somewhere above), it kinda means that the sources **must be** in the lookaside cache already, if we were to just tag the builds for testing/stable.

research/integrations/fedora-ci/index.md
mfocko: We can use Koji scratch build… TBF I'm not sure why Zuul does it that way, but it could be caused by the fact that Zuul is not event-driven, i.e., it's not possible for it to wait for the build to finish and reuse the Koji build. :eyes: That doesn't make sense either… so it does both Koji scratch **and** its own build to give the Testing Farm?

research/integrations/fedora-ci/index.md
mfocko: Maybe we could _abuse_ `packit_instances` to tell apart normal Packit config and what is supposed to be a Fedora CI? With one config for both Packit and “Packit as Fedora CI”, it will eventually become an incomprehensible mess (not to mention syncing of configs from upstream to downstream)

research/integrations/fedora-ci/index.md
mfocko: no, they're not; will reorder

research/integrations/fedora-ci/index.md
mfocko: done

research/integrations/fedora-ci/index.md
lachmanfrantisek: Yes, on the one hand, people want as few configs as possible, but on the other hand, it can easily become a mess.

The idea with instances looks good.

Also, we already have Packit configs in quite a few projects that might be useful with some sane defaults..;) (To not need to go through the full onboarding again.)  Also, as you mentioned elsewhere, ideally not much confing would be needed. So, something to turn this on might be enough.

research/integrations/fedora-ci/index.md
lbarcziova: + maybe also worth considering [this comment](https://github.com/packit/packit/issues/2414#issuecomment-2381490446)

mfocko:
:rocket:

PR #220, opened 2024-12-19
Add global config reasearch

majamassarini:
Related to https://github.com/packit/packit-service/issues/2628

### DIFF
--- a/research/global-config/index.md
+++ b/research/global-config/index.md
@@ -7,7 +7,7 @@ sidebar_position: 1
 
 You can customize the service configuration for a user project in two different ways:
 
-1. There is a configuration on the service side which has to be updated by the service team (upon notification from the service users for opt-in or opt-out from the service). This is how you can enable or disable Zuul for the Fedora CI nowadays: .
+1. There is a configuration on the service side which has to be updated by the service team (upon notification from the service users for opt-in or opt-out from the service). This is how you can enable or disable Zuul for the Fedora CI nowadays.
    I will refer to this solution as a **top-down solution**.
 2. There is a configuration, related with the service, on the user's project side. This is how packit works today. I will refer to this solution as a **bottom-up solution**.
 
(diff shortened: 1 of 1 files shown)


research/global-config/index.md
```diff
@@ -0,0 +1,281 @@
...
+## Jinja2 vs Ansible library
+
+For template management I would probably just use the **jinja2 template library**, even though we can also think about the ansible library.
+The ansible library could let us use `built-in filters and functions` but I don't see use cases for them and as a cons it has a heavier dependency footprint.
+
+# Performances
+
+Splitting the configuration in multiple configuration files will lead obviously to worst performance. Personally I don't see a way to prevent it.
+
+We can limit the number of recursion steps; 3/4 steps are, from my point of view, more than enough. Having a recursion limit will avoid an infinite recursion for malformed configurations.
```
lbarcziova, line 275: good point, we could even start with 1/2 levels

lbarcziova (commented):
thanks Maja for the detailed write-down with the examples! I just quickly went through them, and the approaches you provided look to me all like some templating mechanisms in the end, or? Could you please also cover something like Zuul config for Fedora dist-git ([see](https://pagure.io/fedora-project-config/blob/master/f/resources/fedora-distgits.yaml)), i.e. a config in one place that is edited by users there?

research/global-config/index.md
```diff
@@ -0,0 +1,361 @@
...
+- Quickest implementation; this approach is the same as for the packit-service configuration file.
+
+## Disadvantages
+
+- It is probably easier, for the final user, to look for the service configuration file in its own repo instead of a service repo.
+- Less user engagement due to limited visibility of configuration changes. Even though a configuration migration can be simpler for the service team it could be less explicit for the final user. Being able to change behaviour without the user acknowledging it could not be a good idea. In packit there already is a configuration migration script.
+- The global configuration file could really grow huge and be difficult to maintain both for the service team and for a final user that wants to contribute to it.
+- Lacks ecosystem-specific configuration management. There is no encapsulation for "middle layer" knowledge, no easy way to manage ecosystem configurations. Configurations could be grouped in different nodes with different defaults but there is no easy way, for the service team, to know if the user who is asking to update a configuration has the rights for doing so.
```
lbarcziova, line 38: this is a good point

research/global-config/index.md
```diff
@@ -0,0 +1,361 @@
...
+# Top-Down solution
+
+The service side maintains project-specific behavior configurations.
+
+## Advantages
+
+- Users don't need configuration files in their projects; they might not like the idea of having a service related file in their projects.
+- Configuration migrations are straightforward for the service team; it is just a change in a file that belongs to the team.
+- Highest performance due to direct configuration access; no need to load and pre-process other files.
+- Quickest implementation; this approach is the same as for the packit-service configuration file.
```
lachmanfrantisek, line 31: For this reason, we can use this to enable dist-git CI for packages before it is run by default. (And nothing else.)

research/global-config/index.md
```diff
@@ -0,0 +1,361 @@
...
+- Differs from Packit's current approach, potentially confusing users. This could work mainly for the _Fedora CI_ and could confuse users that use the standard packit configuration for _upstream continuous integration_ and _release synchronization_.
+
+# Bottom-Up solution
+
+Project-side configuration with ability to reference global configurations.
+
+## Advantages
+
+- User-friendly and explicit. Everything is on the user side (more quick for him) and the file will never grow too much because of other project details.
+- Flexible customization through layered configurations and better ecosystem-specific configuration support.
```
lachmanfrantisek, line 48: Big :+1:  for this. It would be really nice to get SIGs help in this. And maybe create a shared space for putting these together.

research/global-config/index.md
```diff
@@ -0,0 +1,361 @@
...
+## Configuration Layers
+
+### Service-Side Layer
+
+The last layer of a configuration chain could be on the service side and it could be referenced in a sort of opt-in mechanism or it could be automatically applied.
+
+### User-Side Layer
+
+At the moment the packit service needs just one configuration file on the user project side both for the **upstream continuous integration** and for the **release synchronization**.
+When implementing the **Fedora CI** (which is a _downstream continuous integration_) a new configuration file could be required or the configuration can be merged with the existing one.
```
lachmanfrantisek, line 66: We just need to be careful not to introduce too many file names because of the forge API issues (but Forgejo might handle these more easily).

research/global-config/index.md
```diff
@@ -0,0 +1,361 @@
...
+
+##### Cons
+
+Poor flexibility, I don't see an easy way to disable the above test job with trigger commit.
+
+Not really explicit, even though we use a placeholder it is harder to recognize the keys that need overriding.
+
+### Implementation
+
+Personally I find the _global config + overlays_ approach the best and in this case we would need to:
```
lachmanfrantisek, line 310: :+1: Looking at the examples (thanks for them), this one looks the most readable.

research/global-config/index.md
lachmanfrantisek: It can easily happen when copy-pasting examples. But we can also save a list of config files through the calls and check for the recursion.

research/global-config/index.md
```diff
@@ -0,0 +1,361 @@
...
+We can limit the number of recursion steps; 3/4 steps are, from my point of view, more than enough. Having a recursion limit will avoid an infinite recursion for malformed configurations.
+
+### packit-service defaults
+
+It could happen that the packit-service config defaults for the "Fedora CI instance" and those for the "Usual instance" (as an example) diverge.
+
+If it happens we could create two _hidden_, _inner_ packit config bases, one per instance, which will always be used for merging any packit config we process; in this way the differences will be grouped explicitly in a single place and we could, probably, enable and disable jobs for one instance or the other (as an example `pull-from-upstream` should not appear in fedora ci packit configuration) just using templating.
+
+<!--
+We already have a configuration file that enables packit we can still count on it.
```
lachmanfrantisek, line 344: Since this allows multi-level approach, we can have something already for the Packit-owned configs. Similarly to your example above.

* packit-base (just to have something shared for all the packages if we need to influence anything. Maybe loaded by default for all the packages?)
* `packit-usecase` (`upstream-ci`, `downstream-ci`, `fedora-automation`), or `packit-instance`

research/global-config/index.md
```diff
@@ -0,0 +1,361 @@
...
+  - downstream ci
+  - sync release
+
+Or we can have just one packit config
+
+- by default one project in distgit with the packit config would be enabled the downstream CI and for this reason we should probably have a key to be able to opt-out from it (both as a user action or as a packit configuration action) we can achive it using the above template mechanism and always apply out default to a user configuration.
+  Or **the CI experience will be anabled if the user refers the related global config**.
+  However if the user configuration is against the will of our configuration we should decide which configuration wins (I would say ours)
+- it should be visible which is the result of applying the user packit config to a packit instance config. So we should probably have a command line or a packit service command to show the user the resulting configuration.
+-->
```
lachmanfrantisek, line 361: With any solution, we should provide an easy way to get an expanded configuration. Locally, but maybe also in the logs with clear references where these external configs came from.

research/global-config/index.md
lachmanfrantisek: Looking at these Packit-owned ones. What do you think about supporting multiple base configs?

This might be handy to combine e.g. type of package (.eg. ecosystem-specific actions) with jobs setup (e.g.upstream ci).
If there is a list of base configs, we can just apply them one by one. But I get this might introduce too much mess into the system.

lachmanfrantisek (approved):
Very nicely written Maja!

lbarcziova:
@majamassarini can you please merge this?

PR #221, opened 2025-01-16
Add research about running operations in target-matching environments

### LINKED ISSUE #219: Research how to provide target-matching environment
There are various issues we are hitting when trying to do tasks on workers targeting another system.

The ideal solution would be to use the exact target instead of having one for all.
In short term, we are going to switch our workers back to Fedora: https://github.com/packit/deployment/issues/613

Spend a while researching the following (and maybe other) options we might want to use:

* Copr / resalloc
  * Already discussed with praiskup, shortly:
    * Either use Copr's own resalloc instance or set up a custom one.
    * Communication via SSH, most of the tooling would need to be implemented.
* Testing Farm
  * Let's chat with TF team to see if we can't reserve an environment there.
* mock
  * [mock in a container](https://rpm-software-management.github.io/mock/#mock-inside-podman-fedora-toolbox-or-docker-container) requires privileged mode which is not possible in our current OpenShift instance
* Container-based solutions
  * Check if containers are enough.
  * We can either enhance the current worker setup and/or use some other service/tool for that.

lbarcziova:
Fixes #219

### DIFF
--- a/research/target-matching-env/index.md
+++ b/research/target-matching-env/index.md
@@ -1,0 +1,81 @@
+---
+title: Running operations in target-matching environments
+authors: lbarczio
+---
+
+For some of the operations run in our service, it would be beneficial
+to run them in target-matching environment, e.g. parsing the specfile placed
+in F41 branch should happen on F41. Currently, we base our images
+on Fedora 41.
+
+## Use cases
+
+- user-configured commands (`actions` in config)
+  - affected jobs: `copr_build`, `propose_downstream`, `pull_from_upstream`
+  - run in on demand pods in `sandbox` (dedicated Openshift project)
+- building SRPM
+  - affected jobs:
+    - `upstream_koji_build` - in `long-running` workers
+    - `copr_build` - in Copr, we just pass the required env as an argument for the API call
+- specfile parsing
+  - for getting version, manipulating changelog, etc. during release syncing
+  - affected jobs: `pull_from_upstream`, `propose_downstream`
+  - run in `long-running` workers
+- potentially new use cases if Packit runs as Fedora dist-git CI
+
+## Possible solutions
+
+### container-based solutions
+
+#### dedicated workers for each Fedora version
+
+- for building SRPM, running release syncing (=> specfile parsing)
+- build separate images for each Fedora version
+- use Celery for task routing; tasks requiring specific environments would be routed accordingly
+- by default, tasks could run on the "main" worker (e.g. currently F41-based).
+- tasks requiring target-matching environments would be refactored (e.g. release syncing for
+  all branches (currently one task) will need to be split)
+- suggestion from review: for better load balancing, we could consider running all the tasks except `process_message` utilising
+  this routing logic, i.e. the tasks would be refactored so that one task runs for specific target and is always routed accordingly from `process_message`
+- potential waste of resources
+- shouldn't introduce any delays
+
+#### sandcastle adjustments
+
+- for running `actions`
+- build separate images for each Fedora version
+- adjust the code running sandcastle to dynamically pass the image reference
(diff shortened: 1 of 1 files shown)


research/target-matching-env/index.md
```diff
@@ -0,0 +1,70 @@
...
+  - run in on demand pods in `sandbox` (dedicated Openshift project)
+- building SRPM
+  - affected jobs: 
+    - `upstream_koji_build` -  in `long-running` workers
+    - `copr_build` - in Copr, we just pass the required env as an argument for the API call
+- specfile parsing
+  - for getting version, manipulating changelog, etc. during release syncing
+  - affected jobs: `pull_from_upstream`, `propose_downstream` 
+  - run in `long-running` workers
+- anything else?
```
lbarcziova, line 22: let me know if you know about some other use-cases I should consider here

research/target-matching-env/index.md
```diff
@@ -0,0 +1,79 @@
...
+
+### container-based solutions
+
+#### dedicated workers for each Fedora version
+
+- for building SRPM, running release syncing (=> specfile parsing)
+- build separate images for each Fedora version
+- use Celery for task routing; tasks requiring specific environments would be routed accordingly
+- by default, tasks would run on the "main" worker (e.g. currently F41-based).
+- tasks requiring target-matching environments would be refactored (e.g. release syncing for
```
majamassarini, line 36: I am wondering if it could be easier and useful, since we will have workers for every needed Fedora and CentOS, to run the `process_messages` task in the worker with the latest release and then just route any other job based on the target? I am not sure if it is better to distribute more the tasks on the workers, since we have them. Or not.

research/target-matching-env/index.md
```diff
@@ -0,0 +1,79 @@
...
+
+- [mock in a container](https://rpm-software-management.github.io/mock/#mock-inside-podman-fedora-toolbox-or-docker-container)
+  requires privileged mode which is not possible in our current OpenShift instance
+
+## Next steps
+
+- with switching to F41 based image, we have unblocked users needing the latest RPM macros
+- once there is more user demand for any of the use cases, we can consider the options above, depending on
+  the particular use-case
+  - the container-based solution with dedicated workers looks like the best fit for running the whole tasks to me
```
majamassarini, line 77: Personally, I agree with this.

majamassarini (approved):

research/target-matching-env/index.md
lbarcziova: you mean e.g. using the build/test-target-overrides? It is a good point to think about, I will note it here, I see issues only with things like storing to the DB (with the current implementation I think that would mean each target would be in a dedicated group model)

majamassarini: I am not sure I understand what you mean by build/test-target-overrides.
I am suggesting to route every job, except `process_messages`, and not only some of them to the "target" worker.  We can choose the target from the `target` key (in `copr_build`, `test`, `koji` jobs ecc.) and other times using the `dist_git_branches` key (for `propose_downstream` as an example).

I see it easier, probably, to balance resources for the workers in this way; like increasing resources for the worker associated with the latest fedora (that I image it is the most used). And maybe decrease resources for the worker associated with the oldest fedora we have... or something like that. But I don't have much arguments for it, just wondering if it makes sense...

I am not sure if this will change something on the database group model, I can't see it.

lbarcziova: > I am not sure I understand what you mean by build/test-target-overrides.

I meant utilising current build/test_target_override implementation (setting these to the one particular target that will be processed in the task) for doing the routing which you describe later. And I mentioned DB group model because if we process each of the targets in a dedicated task, we need to somehow think about when/how the group model is created. But I get your point about balancing out the utilisation of the workers, so will definitely note that!

PR #222, opened 2025-02-05
Add research about migration to FastAPI

lbarcziova:
Fixes packit/packit-service#2699

### DIFF
--- a/research/fast-api/index.md
+++ b/research/fast-api/index.md
@@ -9,7 +9,7 @@ This research aims to evaluate these benefits and outline the key aspects of a p
 
 ## [FastAPI](https://fastapi.tiangolo.com/)
 
-### Advantages:
+### Advantages
 
 - automatic documentation: OpenAPI, Swagger
 - data validation, type hints: Python’s type annotations and [Pydantic](https://docs.pydantic.dev/latest/) for data validation
@@ -19,6 +19,14 @@ This research aims to evaluate these benefits and outline the key aspects of a p
 - good ecosystem, community
 - async support
 - performance
+- filtering, searching, pagination should be more easier to do
+
+### Disadvantages (from our architecture discussion)
+
+- frequent updates, maintenance overhead
+- Pydantic v2 Rust dependency: Pydantic v2 requires a Rust toolchain, which can be difficult to manage
+  on LTS-based distributions. Since we use Fedora-based images in OpenShift and can also pin dependencies,
+  this is unlikely to be a major blocker.
 
 ### Considerations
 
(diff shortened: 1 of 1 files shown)


lbarcziova:
@Venefilyn could you also have a look please?

research/fast-api/index.md
```diff
@@ -0,0 +1,246 @@
...
+
+## Migration approach
+
+### Gradual rewrite
+
+- run Flask and FastAPI together
+- pick a few endpoints (e.g. file by file) and migrate
+- redirect traffic to FastAPI for migrated endpoints
+- repeat until everything is in FastAPI
+- remove Flask completely
```
Venefilyn, line 108: IMO I think keeping the endpoints for Flask is better and then introducing a documented FastAPI under something like `/api/v1/` to make the API documented. Then we wouldn't need to redirect traffic but simply mark it as deprecated until it breaks (or we remove it)

research/fast-api/index.md
```diff
@@ -0,0 +1,246 @@
...
+            project=(
+                ProjectModel(
+                    project_url=build.get_project().project_url,
+                    repo_namespace=build.get_project().namespace,
+                    repo_name=build.get_project().repo_name,
+                )
+                if build.get_project()
+                else None
+            ),
+        )
```
Venefilyn, line 233: This would also be moved or migrated to expected FastAPI way of doing things to make the endpoint self-documenting. Otherwise getting this documented over time is difficult without manual declaration. I think it should be possible in some sort of function wrapper to ensure the document is documented, or by using a pagination library together with this

research/fast-api/index.md
```diff
@@ -0,0 +1,246 @@
...
+- revisit error handling, utilise built-in support
+
+### Implementing models
+
+This is not strictly required (for the GET requests) but significantly improves the migration's value and is highly recommended.
+
+Previous attempt to document the endpoints with Flask:
+
+- https://github.com/packit/packit-service/pull/2089/files
+- the biggest issues: lot of duplication and additional code, change in serialization
```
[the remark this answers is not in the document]
lbarcziova: agreed, I will mention this explicitly there

research/fast-api/index.md
```diff
@@ -0,0 +1,246 @@
...
+
+## [FastAPI](https://fastapi.tiangolo.com/)
+
+### Advantages:
+
+- automatic documentation: OpenAPI, Swagger
+- data validation, type hints: Python’s type annotations and [Pydantic](https://docs.pydantic.dev/latest/) for data validation
+- built-in support for dependency injection, which is useful for managing database
+  sessions, authentication, and configuration
+- websocket and background task support - can be helpful for real-time updates or long-running processes (for usages stats?)
```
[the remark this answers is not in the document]
majamassarini: I agree that this sounds useful for collecting stats outside of our celery workers.

dcermak:
> Am I wrong or are we missing the disadvantages that @dcermak suggested to us on arch meeting?

These are really just two things:
1. larger amount of churn and hence more work to stay up to date
2. pydantic v2 is requiring an up to date Rust toolchain and can be a bit of a pain to get running on "LTS Distros", especially if you want to package this as rpms.

However, I cannot judge whether either of them are relevant to you

lbarcziova:
@majamassarini no, I haven't pushed my updates here yet. But as @dcermak wrote, while acknowledging those, I don't think they (the 2.point mainly) will be affecting our setup too much.

research/fast-api/index.md
```diff
@@ -0,0 +1,253 @@
...
+- async support
+- performance
+- filtering, searching, pagination should be more easier to do
+
+### Disadvantages (from our architecture discussion)
+
+- frequent updates, maintenance overhead
+- Pydantic v2 Rust dependency: Pydantic v2 requires a Rust toolchain, which can be difficult to manage
+  on LTS-based distributions. Since we use Fedora-based images in OpenShift and can also pin dependencies,
+  this is unlikely to be a major blocker.
```
majamassarini, line 29: Thanks

majamassarini (approved):
LGTM!

PR #223, opened 2025-02-06
Add notes for job cancelling

mfocko:
Fixes packit/packit-service#2694

### DIFF
--- a/research/ux/job-cancelling.md
+++ b/research/ux/job-cancelling.md
@@ -123,4 +123,39 @@ cancel running scans.
 
 Suggested splitting into subtasks:
 
-_TODO_
+- [ ] Implement cancelling in the Packit API (knowing what needs to be
+      cancelled, i.e., Copr build ID, Testing Farm request, etc.)
+
+- [ ] Implement methods that would yield respective jobs to be cancelled, i.e.,
+      after retriggering a Copr build, we should get a list of running Copr
+      builds associated with the previous trigger
+
+- [ ] Improve the previous method by incorporating subsequent jobs
+
+  - _NOTE_: this might get more complex after implementation of job dependencies
+
+- [ ] Allow users to cancel running jobs via comment
+
+- [ ] Allow users to cancel running jobs via custom GitHub Check _action_
+
+  - _NOTE_: custom action can incorporate additional metadata provided by us,
+    therefore cancelling this way could be pretty cheap (there would be no need
+    to deduce which jobs need to be cancelled)
+  - _NOTE_: there's a smallish issue of differentiating of what should be
+    cancelled (could be handled by multiple custom actions), for example:
+    - _Copr build for specific target_
+    - _Copr build for all targets_
+    - _Copr build for all targets matching an identifier_
+
+- [ ] Automatically cancel running jobs once an update happens, e.g., push to
+      a PR, branch, or retriggering via comment.
+
+- [ ] (optionally, low-prio) Allow this to be configurable
+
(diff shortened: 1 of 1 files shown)


research/ux/job-cancelling.md
```diff
@@ -0,0 +1,152 @@
...
+### OpenScanHub
+
+Based on a brief look through the docs, it appears that it is not possible to
+cancel running scans.
+
+## Breakdown
+
+Suggested splitting into subtasks:
+
+- [ ] Implement cancelling in the Packit API (knowing what needs to be
```
lbarcziova, line 117: do you think we could start with only the most frequently used jobs (=> most resources), i.e. Copr builds and Testing Farm?

research/ux/job-cancelling.md
```diff
@@ -0,0 +1,152 @@
...
+      cancelled, i.e., Copr build ID, Testing Farm request, etc.)
+
+- [ ] Implement methods that would yield respective jobs to be cancelled, i.e.,
+      after retriggering a Copr build, we should get a list of running Copr builds
+      associated with the previous trigger
+
+- [ ] Improve the previous method by incorporating subsequent jobs
+
+  - _NOTE_: this might get more complex after implementation of job dependencies
```
lbarcziova, line 126: we should note this in the epic/issue for it

research/ux/job-cancelling.md
```diff
@@ -0,0 +1,152 @@
...
+
+- [ ] Allow users to cancel running jobs via comment
+
+- [ ] Allow users to cancel running jobs via custom GitHub Check _action_
```
lbarcziova, line 130: these are definitely nice to have, but the cancelling after retriggering/new push etc. (the point below) is in my opinion the most crucial at the moment

research/ux/job-cancelling.md
mfocko: I don't think it would be a lot of work to add the support for cancelling on our end… Koji seems to be the most complex out of all, the most tedious part is the deduction of what needs to be cancelled (on the service-side)

research/ux/job-cancelling.md
majamassarini: Even though I agree with this note, I think, on the other hand, that now that we don't have job dependency it will be more complicated to cancel a test job that depends on a copr build cancelled by a user.

research/ux/job-cancelling.md
```diff
@@ -0,0 +1,222 @@
...
+Suggested splitting into subtasks:
+
+- [ ] Implement cancelling in the Packit API (knowing what needs to be
+      cancelled, i.e., Copr build ID, Testing Farm request, etc.)
+
+- [ ] Implement methods that would yield respective jobs to be cancelled, i.e.,
+      after retriggering a Copr build, we should get a list of running Copr
+      builds associated with the previous trigger
+
+- [ ] Automatically cancel running jobs once an update happens, e.g., push to
```
lbarcziova, line 194: we could maybe mark this as the hard requirement for the epic for this quarter to have this, and the following to be follow-ups (coming back to https://github.com/packit/research/pull/223/files#r1946075678)

lbarcziova (approved):
thanks, nicely done!

research/ux/job-cancelling.md
mfocko: I guess it's accounted for in the [tracking issue](https://github.com/packit/packit-service/issues/5#issue-447593326), everything below this point is under “improvements”

lbarcziova: ah sorry, missed that, great!

PR #231, opened 2026-05-19
MR consolidation research

majamassarini:


research/ymir-issues-consolidation/index.md
```diff
@@ -0,0 +1,436 @@
...
+
+**Parent Issue (Plan E)**: Create parent Jira issue to orchestrate and track consolidation with maintainer approval.
+
+**Hybrid (Plan D)**: Combine multiple approaches with per-package configuration policy.
+
+### Comparison Table
+
+| Approach                      | Effort    | Flexibility         | Unblocking    | Composable | Best For         | Complexity |
+| ----------------------------- | --------- | ------------------- | ------------- | ---------- | ---------------- | ---------- |
+| **Incremental (RECOMMENDED)** | **Low**   | ✅ High             | ✅ Easy skip  | ✅ Yes     | **2-4 issues**   | **Low**    |
```
TomasTomecek, line 61: agreed, this indeed looks like a great design with immense flexibility

research/ymir-issues-consolidation/index.md
```diff
@@ -0,0 +1,436 @@
...
+
+- Incremental consolidation is **composable** - can add orchestrator later (Phase 2: Low effort) for automated batch-like experience. Total effort remains Low-Medium for full automation, but phased deployment reduces risk.
+- Supports **cross-type consolidation** - backport and rebuild issues for the same package+fixVersion can be consolidated into a single MR.
+
+### Recommended Solution: Incremental Consolidation
+
+**Use two Jira labels to enable step-by-step consolidation:**
+
+- **`consolidate_base`** - Marks the current "base" issue (whose MR branch others build on)
+- **`consolidate_next`** - Marks the next issue to consolidate on top of base
```
TomasTomecek, line 78: one slight concern: people would forget this happens in jira and would apply labels in gitlab; we should improve the MR description and guide them there that if they wanna include another fix in the same MR they need to do steps a b c

research/ymir-issues-consolidation/index.md
```diff
@@ -0,0 +1,436 @@
...
+
+**How it works:**
+
+1. RHEL-1000 → normal backport → MR1 created
+2. Mark RHEL-1000 with `consolidate_base`
+3. Mark RHEL-2000 with `consolidate_next`
+4. Backport agent forks from MR1's branch (not c9s), applies patch
```
TomasTomecek, line 85: doesn't even need to fork, just check it out (and maybe rebase just in case?)

research/ymir-issues-consolidation/index.md
```diff
@@ -0,0 +1,436 @@
...
+5. MR2 created (contains both patches), MR1 closed, base label moves to RHEL-2000
```
TomasTomecek, line 86: actually, this is wasteful right? we can reuse that MR and just update title & description

research/ymir-issues-consolidation/index.md
```diff
@@ -0,0 +1,436 @@
...
+#### 1. Modify Backport and Rebuild Agents
+
+**Location**: `ymir/agents/backport.py` and `ymir/agents/rebuild.py`
+
+**Key changes (apply to both agents):**
+
+1. **Detect consolidation mode**: Check for `consolidate_next` label
+2. **Find base branch**: If triggered, search for issue with `consolidate_base` label (same package+fixVersion)
+3. **Fork from base**: Use base MR's branch instead of target branch (c9s/c8s)
+4. **Validate base MR**: Must be opened, branch must exist
```
TomasTomecek, line 168: I wonder if the agent should also read all comments and discussions from the MR and incorporate them

TomasTomecek (approved):
thank you, very nice research! I stopped reading at 50% because it was getting way too long, sorry 😊

research/ymir-issues-consolidation/index.md
lbarcziova: 2 notes to this:
1. we might want to prefix these labels too, if this is Ymir-specific behaviour, so e.g. `ymir_consolidate_base` 
2. did you compare the labels on PR vs on Jira approach? I would be curious about that, I assume both has advantages, but to Tomas note, to me, it also looks like a more natural place (might be totally biased, we could try to get more opinions from maintainers on this)

research/ymir-issues-consolidation/index.md
```diff
@@ -0,0 +1,436 @@
...
+
+1. RHEL-1000 → normal backport → MR1 created
+2. Mark RHEL-1000 with `consolidate_base`
+3. Mark RHEL-2000 with `consolidate_next`
+4. Backport agent forks from MR1's branch (not c9s), applies patch
+5. MR2 created (contains both patches), MR1 closed, base label moves to RHEL-2000
+6. Repeat: mark RHEL-3000 with `consolidate_next`, it builds on MR2
+7. Final MR contains all consolidated patches
+
+**Cross-type consolidation (backport + rebuild):**
```
lbarcziova, line 90: do you recommend, going with approach 1, to still keep the current behaviour for rebuilds (consolidation done here always), or do you have any suggestions how to improve this? One thing I'm not sure about is that we don't consolidate issues already labeled with `ymir_rebuilt`, I implemented it this way to avoid race conditions and Jira triggering on multiple issues to create multiple duplicated issues. (cc @opohorel)

lbarcziova (commented):
thanks Maja! I agree to start iterating with approach 1 and decide based on it where to go with it.

research/ymir-issues-consolidation/index.md
majamassarini: Honestly, I told Claude I want labels on Jira issues, and I emphasized this because I think it provides a simpler UX. All the labels so far are applied to Jira issues, so I'd stick with that, it's easier for users to remember, and we have a single place with all the Ymir-call history. But we can discuss it more.

research/ymir-issues-consolidation/index.md
majamassarini: I am not sure 🤔 If the agent fails (in a non obvious way) to create the backport for the second issue, I want to be able to move to the third one. I don't want to touch the MR created in the first run. By 'check it out,' do you mean `check it out and create a new branch on top of it`? If so, then yes. I agree, 'forks' is misleading here.

research/ymir-issues-consolidation/index.md
majamassarini: I wouldn't reuse the same MR for the same reason as in the previous point. But we can discuss it on arch. In my mind, I would like to be able to try different consolidation paths (for example different backport approaches) and for that I need to keep all the intermediate MRs 'standalone'. So that it is easy to reuse them if needed.

research/ymir-issues-consolidation/index.md
majamassarini: It could probably be useful for solving potential conflicts. I can add it, but I'd keep this as a follow-up (if/when needed). If we don't have evidence, I wouldn't add all that context.

research/ymir-issues-consolidation/index.md
majamassarini: Yes, I would still keep the current behaviour for rebuilds consolidation. It already works and automatically consolidates multiple issues at once. Its output can still be consolidated into a backport later. In this case we would be consolidating an issue already labeled with `ymir_rebuilt`, but that seems fine to me. I wouldn't force skipping the issues with `ymir_rebuild` label.

research/ymir-issues-consolidation/index.md
TomasTomecek: yes, let's discuss, I definitely have opinions and arguments here :)

research/ymir-issues-consolidation/index.md
TomasTomecek: agreed to do that as a followup

research/ymir-issues-consolidation/index.md
majamassarini: We agreed on arch meeting to keep the labels in Jira, also because the `ymir_consolidate_next` label can't be put in a non yet existing MR.

research/ymir-issues-consolidation/index.md
majamassarini: We agreed on arch meeting to start with keeping all intermediate MR which is the safest solution if the "next step" introduces a failure for example in the CI (we can change this behavior later, probably quite easily, if it bothers maintainers too much).

research/ymir-issues-consolidation/index.md
```diff
@@ -1,5 +1,39 @@
...
+
+**Problem**: A single upstream patch addresses multiple CVEs at once, but each CVE has a separate downstream Jira issue. Without consolidation, the agent would create duplicate backport MRs for the same patch.
+
+**Example**:
+
+- Upstream commit fixes CVE-2024-1234, CVE-2024-5678, CVE-2024-9012 in one patch
+- Downstream has RHEL-1000, RHEL-2000, RHEL-3000 (one per CVE)
+- Agent triggered on RHEL-1000 should detect RHEL-2000 and RHEL-3000 are siblings
+
+**Solution**: Automatic triage-time detection that parses upstream commit message, finds related downstream issues, and consolidates into single MR with `Resolves: RHEL-1000, RHEL-2000, RHEL-3000`.
```
antbob, line 31: The solution would also need to include Jira issue updates for each separate CVE (or non CVE issue for that matter) issue:

1. transition each Jira issue to "In Progress" as it is required by the downstream automation.
2. add a comment to each Jira issue indicating is part of the backport OR link to related MR.

Under (1) once MR is merged then all related Jira issues will have fixedInBuild field updated and Preliminary Testing set to Requested by the downstream automation pipeline and then once  Preliminary Testing is Pass the downstream automation pipeline will transition each issue to Integration and attach it to new or existing errata.

majamassarini: I will add both points to the doc, thanks!
The rebuild consolidation already comments on all consolidated jira issues, but it does not change the status to "In Progress". @lbarcziova do you think we need to change the behaviour also in the rebuild consolidation? Or should we diverge in this aspect?

lbarcziova: we can do that to be consistent with backport/rebase

PR #233, opened 2026-05-20
Research async building architecture

nforro:


majamassarini (approved):
From the proposed solutions, personally I would start with option A (async multiplexing).

However, I am not totally sure we should disregard the idea of relying on the CI builds.
We would like to rely on the CI tests in the future, right? So why not also on the CI builds?
If we mark the MR as incomplete (WIP/draft), I think we can iterate on it multiple times based on the CI feedback. The iteration loop would be almost the same for the builds and for all the other CI checks. Also, we could collect "intermediate feedback from maintainers" in this way, something that other solutions will not give us.
I think it really depends on how many times we usually need to iterate on a build. And, I may be wrong, but as we improve things, I think we are iterating less and less on it.

research/ymir-async-building/index.md
```diff
@@ -0,0 +1,421 @@
...
+    if element:
+        _, payload = element
+        t = asyncio.create_task(process_task(payload))
+        active.add(t)
+        t.add_done_callback(active.discard)
+```
+
+**Advantages:**
+
+- Minimal code changes — workflows are already async
```
TomasTomecek, line 199: really love that, no need for celery, just a few code changes

research/ymir-async-building/index.md
```diff
@@ -0,0 +1,421 @@
...
+- Copr polling already uses `asyncio.sleep`, yielding the event loop — this
+  includes builds triggered inside the cherry-pick fix agent's LLM session,
+  since the `build_package` tool call goes through the same async polling loop
+- Single pod, single process — simpler operations
+
+**Challenges:**
```
TomasTomecek, line 205: how about logs? how hard will it be for us to diagnose issues from logs?

research/ymir-async-building/index.md
```diff
@@ -0,0 +1,421 @@
...
+  conversation. Source processing is the dominant consumer — large packages
+  (Firefox, Thunderbird) need ~4Gi just for unpacked sources. With N=2, a
+  backport pod could need ~12Gi; with N=3, ~18Gi. The SE deployment already runs backport pods at 6Gi limits for single
+  tasks.
+- **MCP Gateway**: the MCP protocol multiplexes requests by ID over a single
+  SSE session (`_response_streams` dict keyed by `RequestId`), so concurrent
+  tool calls from the same pod work out of the box. The gateway tools
+  (e.g. `BuildPackageTool`) are stateless. No gateway changes needed.
+- **Error blast radius**: one task's crash could affect others in the same
+  process.
```
TomasTomecek, line 217: it would be great to know if we can prepare the codebase for this change

research/ymir-async-building/index.md
nforro: Good point, we would probably have to augment the log messages to be able to differentiate between tasks, similarly to packit-service with Celery.

research/ymir-async-building/index.md
nforro: Response from Claude:

  Yes, several incremental prep steps would make Option A easier without committing to it yet:

  - Add a task/correlation ID to the workflow context and thread it through all log messages (using Python's logging filter or contextvars). This is useful regardless of concurrency — it improves traceability even with single-task pods.
  - Audit shared mutable state — check if any module-level globals, singletons, or shared file paths (e.g. /git-repos/{issue}/) would collide when two tasks run concurrently. The PVC paths already use {issue} as a namespace, which is good.
  - Ensure MCP gateway session isolation — verify that the _response_streams dict keyed by RequestId truly handles concurrent tool calls without cross-talk.
  - Profile actual RSS on a representative set of packages during source processing. This determines whether N=2 is feasible without a quota bump.

research/ymir-async-building/index.md
```diff
@@ -0,0 +1,421 @@
...
+  the constructors already accept dicts internally. BeeAI also offers
+  alternative memory classes (`SlidingMemory`, `TokenMemory`,
+  `SummarizeMemory`) that could help bound conversation size for long-running
+  fix loops.
+- Each `fix_build_error` call creates a **fresh agent** with no memory of prior
+  fix attempts — context is passed entirely through the prompt template (build
+  error message, upstream patches, repo path). This means conversation
+  continuity across the build wait is **not required**.
+
+## Alternative: Skip Build Verification Entirely
```
lbarcziova, line 135: while this would simplify things a lot, I am a bit skeptical - I remember agent sometimes iterated many times on build failures during pilot, and re-pushing 5-10 times to a draft MR might be quite noisy. On the other hand, this might be statistically happening in minority of cases (and maybe more common in rebases than in backports?) But definitely worth validating against recent runs, @opohorel @abobrov any data from your experience on this?

research/ymir-async-building/index.md
```diff
@@ -0,0 +1,421 @@
...
+## Open Questions
+
+1. **What is the actual RSS of a backport pod during source processing?**
+   Feedback from the SE deployment shows large packages (Firefox, Thunderbird)
+   need ~4Gi just for source extraction, making even N=2 concurrency
+   challenging. Profiling smaller packages would clarify the typical case.
+2. **What's the namespace quota expansion path?** Determines feasibility of
+   Option A at higher concurrency.
+3. **How large are typical agent conversations?** Affects state size
+   estimates and memory profiling for Option A.
```
lbarcziova, line 421: @nforro would you create a POC issue that we can prioritise and decide which option to go with?

lbarcziova (approved):

research/ymir-async-building/index.md
opohorel: I would advise against skipping the build verification. quite a few backports are not passing through it, forcing the agent to improve the backport. although I saw that the number of consecutive build failures decreasing with newer models, but we are not there yet

research/ymir-async-building/index.md
TomasTomecek: makes sense, we can prepend with some session id or ideally even with human readable identification, something like "f{jira}-{branch}-{attempt}"

research/ymir-async-building/index.md
TomasTomecek: Claude: implement it and propose a PR :sweat_smile:

PR #235, opened 2026-06-24
Add postponed CVE issues reprocessing research

majamassarini:
Document mechanism for periodically reprocessing postponed CVE issues in Ymir. Includes 3-phase implementation approach: basic sweeps, backoff optimization with Redis, and follow-up improvements.

### DIFF
--- a/research/postponed-issues-reprocessing/index.md
+++ b/research/postponed-issues-reprocessing/index.md
@@ -74,12 +74,13 @@ Issues can change postponement reasons:
 **Base frequency (Phase 1):** Every 6 hours, checks all postponed dependency issues
 
 **Backoff policy (Phase 2):**
+
 | Sweep attempts | Interval |
-|----------------|----------|
-| 1-3 attempts | 2 hours |
-| 4-8 attempts | 6 hours |
-| 9-15 attempts | 12 hours |
-| 16+ attempts | 24 hours |
+| -------------- | -------- |
+| 1-3 attempts   | 2 hours  |
+| 4-8 attempts   | 6 hours  |
+| 9-15 attempts  | 12 hours |
+| 16+ attempts   | 24 hours |
 
 **Rationale:** Cheap operation, but dependencies may take days/weeks to fix. Aggressive backoff (Phase 2) prevents wasteful checks on long-blocked issues.
 
@@ -96,11 +97,12 @@ Issues can change postponement reasons:
 **Base frequency (Phase 1):** Every 12 hours
 
 **Backoff policy (Phase 2):**
+
 | Sweep attempts | Interval |
-|----------------|----------|
-| 1-4 attempts | 6 hours |
-| 5-10 attempts | 12 hours |
-| 11+ attempts | 24 hours |
+| -------------- | -------- |
+| 1-4 attempts   | 6 hours  |
+| 5-10 attempts  | 12 hours |
+| 11+ attempts   | 24 hours |
 
 **Rationale:** Z-stream errata ship on a predictable but slow cadence. Backoff (Phase 2) moderates checks for long-pending errata.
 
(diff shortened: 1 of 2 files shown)


research/postponed-issues-reprocessing/index.md
```diff
@@ -0,0 +1,479 @@
...
+Jira labels are used as the primary tracking mechanism because they provide efficient JQL queries, require no Jira admin approval, and are already familiar to the team/users. Alternative solutions considered include custom Jira fields (higher API load, requires admin setup), external PostgreSQL database (durable but adds operational complexity), and label swapping for audit trail (fragile, relies on undocumented Jira behavior). The chosen approach uses labels for state, Redis for sweep optimization, and brief comments for human-readable context, balancing simplicity with scalability.
+
+### Labels as Primary Signal
+
+Each postponement reason has a corresponding label:
+
+- `ymir_postponed_dependency` — waiting for a dependency CVE to be fixed
+- `ymir_postponed_y_stream` — waiting for Z-stream errata to ship
+- `ymir_postponed_no_patch` — no upstream patch found yet
+- `ymir_postponed_pr_pending` — patch identified but not yet merged
```
lbarcziova: I'm wondering if we should treat this the same, in case the pending PR would get closed and other approach would be chosen.

research/postponed-issues-reprocessing/index.md
```diff
@@ -0,0 +1,479 @@
...
+**Transition logic:**
+
+- Remove old postponement label
+- Add new postponement label (if still blocked)
+- Update comment with new context
+- Reset backoff tracking (if using Redis in Phase 2)
+
+## Sweep Types and Timing
+
+### 1. Dependency CVE Sweep
```
lbarcziova, line 60: for this one, we actually added the check if the new build is present in compose, I think we should be checking also that one now (if fixed in build is set), could you please re-consider it with that? Should be just different programmatic check

research/postponed-issues-reprocessing/index.md
```diff
@@ -0,0 +1,479 @@
...
+**Action on finding patch:**
+
+- If patch is merged: remove postponement label, push to triage queue
+- If patch exists but not merged: transition to `ymir_postponed_pr_pending`, add PR URL comment
+
+---
+
+## Backoff Optimization with Redis (Optional - Phase 2)
+
+**Why backoff is needed:** Without backoff, all postponed issues are checked on every sweep, even if recently verified as still blocked. This wastes API calls and agent tokens, especially for issues blocked for weeks.
```
lbarcziova: I'm not fully convinced the backoff is necessary/wanted here, I think with most of these there is not much predictability on when we get unblcoked. For the expensive case (rerunning the triage agent) a longer fixed interval would be simpler. But open to discussion. Definitely would start with phase 1 only as you propose!

lbarcziova (approved):
thanks for a comprehensive research, looking forward to start working on this!

research/postponed-issues-reprocessing/index.md
TomasTomecek: agreed, sounds like we should do a full triage once we are unblocked

research/postponed-issues-reprocessing/index.md
```diff
@@ -0,0 +1,479 @@
...
+- Historical trends (growth/shrinkage)
+- Distribution by component
+
+**What Jira dashboards cannot show:**
+
+- Sweep effectiveness (unblock rate per sweep run)
+- Time-to-unblock (how long issues stay postponed)
+- Backoff behavior (attempt count distribution)
+
+### Redis Metrics (Optional - Phase 2 only)
```
TomasTomecek, line 293: really love this part that we'll be able to measure these issues and act based on real data

research/postponed-issues-reprocessing/index.md
majamassarini: I agree the unblock action will be the same, however I would keep them separate because `ymir_postponed_no_patch` requires a full re-triage every time we check it, so I would re-run it less frequently.
The  `ymir_postponed_pr_pending`, instead, needs a github/gitlab check - to verify if the PR (which will be linked in the issue by the triage agent) has been approved, so we can run it more frequently. If the PR has been merged then we can invoke the triage agent again.

And yes, I agree that if PR is closed then we should consider falling back to triage.

research/postponed-issues-reprocessing/index.md
```diff
@@ -0,0 +1,487 @@
...
+
+**Action on unblock:** Remove postponement label, trigger full re-triage from scratch.
+
+---
+
+### 2. Y-stream CVE Sweep
+
+**What it checks:** Whether the Z-stream build is present in the target buildroot/compose
+
+**Method:** Use `check_build_in_buildroot()` (from `ymir/common/utils.py`) to verify the "Fixed in build" NVR from the Z-stream blocker issue is available in the Y-stream buildroot
```
lbarcziova, line 98: I think this should be moved to the 1., for 2. the errata check should be enough

